PR #20, opened 2024-09-21
Check that all variables are defined in the config

Gerrrr:
This change makes configuration parsing throw if at least one environmental variable is not set. This completely eliminates configuration hustle where you are not sure if all required variables are passed correctly.


smccarthy788 (approved):
Very nice, thank you!

PR #27, opened 2025-01-06
Upstreaming 2024 Nyrkiö patches

henrikingo:


henrikingo:
Ah. I think I know what this is: We use hunter together with Nyrkiö, and the latter has pytest-benchmark, so it successfully went through all our own testing. I'll addpytest-benchmark to poetry files and try again. (But not today.)

Gerrrr:
All the changes look good! We just need to remove `pytest-benchmark ` or update the lock.

henrikingo:
Ok actually I think you need to restart the test. I'm not a committer yet :-D

henrikingo:
> Ok actually I think you need to restart the test. I'm not a committer yet :-D

Never mind. It was just so fast I didn't realize it had ran already.

smccarthy788 (approved):

PR #35, opened 2025-02-25
remove unsupported comment

brucearctor:
It isn't great for community or potentially onboarding users for the top of the README to say is unsupported.  So, removed/changed that line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Hunter – Hunts Performance Regressions
 ======================================
 
-_This is open source project was created by DataStax employees._
+_This open source project was created by DataStax employees._
 
 
 Hunter performs statistical analysis of performance test results stored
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 Hunter – Hunts Performance Regressions
 ======================================
 
-_This is an unsupported open source project created by DataStax employees._
+_This is open source project was created by DataStax employees._
```
Gerrrr, line 4: nit: `_This open source project was created by DataStax employees._`

brucearctor: updated

henrikingo: I would remove this entire line in the same PR that we will need to do to change the license headers in every file. Apparently this needs to be done by a Datastax employee. (As Datastax is the donor for the code.)

PR #36, opened 2025-02-25
Updated python

brucearctor:
To have a non-EOL python version


Gerrrr:
Hey @brucearctor ,

Thanks for starting this work! Is it possible to keep compatibility with 3.8 / run test matrix against both versions? 

As I mentioned on the mailing list thread about this topic, hunter/otava is the low-maintenance set-and-forget kind of software that just works and, as a result, may run on old distros/python versions. It would help with migration of existing user base to the Apache version if at least the first release can be switched over without additional hustle of updating distro and/or Python version, etc.

What do you think?

brucearctor:
still [ obviously ] a work in progress [ so have marked as draft ] -- using as a way to get more deeply into the code Was trying to hack around a bit.

Not opposed to maintaining compatability, BUT, with some of the bumps in versions, this might not be possible [ or at least easy ].

brucearctor:
looks like this needs to not be a 'draft' for CI to run.

I'm currently just using that for testing, rather than setting up locally.  

Will tag individuals once ready for review [ ideally with everything passing ].

brucearctor:
I'm gonna close ...  I see this taking quite some time.  AND, I am not ready to devote a bunch of time to get this to work [ when I'm not even a user of it, yet ].  Gonna play with some other stuff for a bit.  Will be happy to follow along, and maybe get this worked out in the future.

PR #37, opened 2025-03-14
Rename Hunter to Otava and add LICENSE headers

### LINKED ISSUE #30: Update Readme to clarify support status?
the ReadME has "This is an unsupported open source project created by DataStax employees." has that at the top.

Is that accurate?

Perhaps this was created by DataStax employees is correct.  BUT, as an 'active' Apache [ incubating ] project, is this really not supported?  Highlighting an open source project is not supported seems like a good way to turn users/contributors away!?

Thoughts?  What's the status here?  Should that wording be updated/removed?

### LINKED ISSUE #31: Adding Apache License Headers
When this mass change's PR is readied the commit should come from a PPMC member who is from the org that signed the SGA.

smccarthy788:
Closes #30. Closes #31.

The main criteria I used to choose how to automate the license header enforcement was integrating into our existing CI built around `poetryw` & `toxw`. In particular I didn't want to introduce downloading some binary or using another language to run a tool. Turns out that was a lot to ask for... I didn't find any good options until I came across how this is done in [Apache Airflow](https://github.com/apache/airflow), using `pre-commit`.

Still... IMO commit hooks are kind of clunky, I'm in no way wedded to this approach. At the very least: the license headers have been added.

The license headers can be applied to new files by running the normal formatting command:
```
./toxw -e format
```
... and being part of the tox testenv will also be tested in CI. There are some extra steps to update the license header, documented in `CONTRIBUTING.md`.


henrikingo:
Wow,thanks Sean. Cool that you also changed the name. This is a big step in completing the transition to ASF.

Given the sweeping nature of this patch I would be inclined to merge it and I'm therefore +1 it herẹ. I have some comments, but IMO we can address those incrementally in follow up work.

I like that checking for license headers is integrated to CI and just generally completely automated. Well done.

+1

====
Comments for discussion and we can work on this in the coming weeks:

I feel like putting the ASF license header into config files is going a bit too far. IMO config files can typically be considered not creative work and therefore not copyrightable. TheASF Source Header and Copyright Notice Policy seems to recognize this in https://www.apache.org/legal/src-headers.html#faq-exceptions

Second, I would also like to find out whether there is any similar  leeway for our docs in markdown format. Documentation of course is definately copyrightable. My aversion to the license headers added here is more that Markdown is designed to be readable as it is, without needing to compile it to HTML or anything else. Hence they are both source files and final files and it's a bit subjective which aspect of that you see first. For me they are readable files first.

It seems both "Short informational text files; for example README, INSTALL files. The expectation is that these files make it obvious which product they relate to." and "'Snippet' files that are included in a larger file, when the larger file would have duplicate licensing headers."

At least if we were to build a single PDF or single HTML page that is distributed as our manual, then that would clearly fall into the snippet case.

If we think of docs/*.md as a publication that is already in readable format, then I would say docs/ is the copyrightable work, and the files in the directory are snippets. As an example, the license header is 11 lines, and for example BIG_QUERY.md is 24, so the license header is almost 50% of the content it is protecting.

Furthermore, if we think of the markdown files as readable in themselves, then such a publication should claim its copyright as actual text. Like it would on the data page of a book, for example. So we could have one instance of the ASF license header after the table of contents, or as a separate file, linked in the table of contents.

Third, there are sentences in the text that say, for example, "hunts for performance regressions". It's IMO good that this patch doesn't go beyond the mechanical rename, but of course it makes sense to follow up with a more editorial PR.

henrikingo (approved):

Gerrrr (approved):
This looks awesome, thank you @smccarthy788 !

PR #53, opened 2025-04-17
Add ASF header to .gitignore

Gerrrr:


henrikingo (approved):
I guess I need to pick my battles...

PR #58, opened 2025-05-09
Rename project name to apache-otava

Gerrrr:
This change addresses RC feedback that project names typically contain prefix "apache-". For example, https://github.com/apache/airflow/blob/40789d677972a3d7d0827e5a12329879487f7926/pyproject.toml#L33


henrikingo (approved):
This changes both the tar.gz and whl files, right?

Gerrrr:
Yes!

```
$ ls dist
apache_otava-0.6.0+incubating-py3-none-any.whl  apache_otava-0.6.0+incubating.tar.gz
```

PR #59, opened 2025-06-05
Remove +incubating, add it to description and authors instead

Gerrrr:


pyproject.toml
```diff
@@ -17,11 +17,11 @@
 
 [tool.poetry]
 name = "apache-otava"
-version = "0.6.0+incubating"
-description = "Change Detection for Continuous Performance Engineering"
-authors = ["Apache Otava Developers <dev@otava.apache.org>"]
+version = "0.6.0"
+description = "Apache Otava (Incubating): Change Detection for Continuous Performance Engineering"
+authors = ["Apache Otava (Incubating) Developers <dev@otava.apache.org>"]
```
henrikingo, line 22: I would drop "Developers" but it works like this too

pyproject.toml
```diff
@@ -17,11 +17,11 @@
...
 include = ["DISCLAIMER"]
-packages = [{include = "otava"}]
+packages = [{ include = "otava" }]
```
henrikingo, line 24: Since we're here... what does this do and why is this not apache-otava?

PR #68, opened 2025-07-03
Install instructions use PyPI instead of GH

### LINKED ISSUE #56: Should install be based on a release as opposed to latest git state?
https://github.com/apache/otava/blob/5502d6b8af59cd1640b8639e22b939e24aa32d8f/docs/INSTALL.md?plain=1#L29

I am not an expert on pip but this appears to support that pip will get the code from git as opposed to from an approved release.

Gerrrr:


henrikingo:
Approved, except that I'm unsure whether this fixes #56? I'm starting to get the feeling ASF procedure requires us to have a page in documentation which explains installation from the official release, which is the source tarball. What you do here is of course what every user will use in reality. But please check with the people active on #56 whether this is what they had in mind, or the source release?

PR #106, opened 2025-12-01
Add e2e Graphite test

Gerrrr:


tests/graphite_e2e_test.py
```diff
@@ -0,0 +1,231 @@
...
+        # Wait a bit more for Graphite to fully initialize
+        time.sleep(5)
+
+        # Seed data into Graphite using the same pattern as datagen.sh
+        _seed_graphite_data(carbon_port)
+
+        # Wait for data to be written and available
+        time.sleep(5)
+
+        yield http_port
```
henrikingo: The postgresql version of this yields also the container ID. It bothers me that these are essentially the same function yet they are inconsistent with their return/yield values. Second, why isn't the carbon port returned?

PR #107, opened 2025-12-05
Fixup postgres_e2e_test.py

Gerrrr:
This PR fixes broken master - https://github.com/apache/otava/actions/runs/19937815300/job/57167461866.


henrikingo:
At this point we can and should drop python 3.8 and 3.9 from master.

henrikingo:
With the upgraded python versions, this PR is obsolete right?

PR #117, opened 2025-12-18
analysis.split sorts change points before getting intervals

Gerrrr:
Prior to this patch, change points discovered while scanning sliding
windows are added to the end of the list. At the same time,
SignificanceTester.get_intervals expects change points to be sorted.

As a result, intervals contain 1 or more `None` that cause `ValueError` in
tester.change_point.

The second commit adds perf tests to the CI suite, so we catch this kind of bugs during PR review.


Sowiks:
Went over the code changes, it's an approval from me.
Good catch!

PR #141, opened 2026-03-25
Fix orig edivisive flow

vishnuchalla:
### Description
Fix orig edivsive flow. I am running into below error if more than 10 datapoints are being used for analysis
```
vchalla@vchalla-thinkpadp1gen2:~/myforks/otava$ otava --config-file /tmp/otava-10.yaml analyze tiny10 --last 15 -P 0.001 --window 15 -M 0.0 --orig-edivisive=true
INFO: Computing change points for test tiny10...
Traceback (most recent call last):
  File "/home/vchalla/.local/bin/otava", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/vchalla/myforks/otava/otava/main.py", line 651, in main
    script_main()
  File "/home/vchalla/myforks/otava/otava/main.py", line 578, in script_main
    analyzed_series = otava.analyze(
                      ^^^^^^^^^^^^^^
  File "/home/vchalla/myforks/otava/otava/main.py", line 125, in analyze
    produced_report = report.produce_report(test.name, report_type)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vchalla/myforks/otava/otava/report.py", line 52, in produce_report
    return self.__format_log_annotated(test_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vchalla/myforks/otava/otava/report.py", line 87, in __format_log_annotated
    change = [c for c in cp.changes if c.metric == col_name]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vchalla/myforks/otava/otava/report.py", line 87, in <listcomp>
    change = [c for c in cp.changes if c.metric == col_name]
                                       ^^^^^^^^
AttributeError: 'ChangePoint' object has no attribute 'metric'
```

### Testing
Tested and verified in local
```
vchalla@vchalla-thinkpadp1gen2:~/myforks/otava$ otava --config-file /tmp/otava-15.yaml analyze tiny15 -P 0.001 --window 50 -M 0.0 --orig-edivisive
INFO: Computing change points for test tiny15...
time                       commit      metric1
-------------------------  --------  ---------
2026-01-01 00:00:00 +0000  c1              100
2026-01-02 00:00:00 +0000  c2              101
2026-01-03 00:00:00 +0000  c3               99
[...]

PR #151, opened 2026-05-04
Rename the two different ChangePoint classes for clarity

henrikingo:
The recent work in #96 to replace the original external dependency on the so called "signal processing" repository with our own implementation, introduced new classess ChangePoint and CandidateChangePoint, in change_point_divisive/base.py but also left in place the original ChangePoint class in analysis.py. These come together in series.py, where the newer is renamed as _ChangePoint() and also acts as a parent to older class, thus aligning their signature as much as possible.

It turns out having two similarly named classes can be a source of confusion and bugs. For example, in #141  vishnuchalla fixes a bug that is due to this and has essentially blocked the --orig-edivisive code path completely.

This patch is an effort to make the existence of two separate classes very explicit, by renaming them to ChagePointHunter and ChangePointOtava based on their "lineage".

A test case is added to exercise the --orig-edivisive code path. The test fails, as predicted by #141. The test is now cmmented out. The bug is due to a missing cp.metric property in one variation of the ChangePoint class.

Note that this patch is intended more for discussion than to merge.

### DIFF
--- a/otava/series.py
+++ b/otava/series.py
@@ -27,7 +27,7 @@ from otava.analysis import (
     compute_change_points_orig,
     fill_missing,
 )
-from otava.change_point_divisive.base import ChangePoint as _ChangePoint
+from otava.change_point_divisive.base import ChangePointOtava
 
 
 @dataclass
@@ -72,7 +72,7 @@ class Metric:
 
 
 @dataclass
-class ChangePoint(_ChangePoint[TTestStats]):
+class ChangePointHunter(ChangePointOtava[TTestStats]):
     """A change-point for a single metric"""
     metric: str
     time: int
@@ -140,7 +140,7 @@ class ChangePointGroup:
     prev_time: int
     attributes: Dict[str, str]
     prev_attributes: Dict[str, str]
-    changes: List[ChangePoint]
+    changes: List[ChangePointHunter]
 
     def to_json(self, rounded=False):
         return {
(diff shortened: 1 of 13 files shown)


otava/series.py
```diff
@@ -72,7 +72,7 @@ def to_json(self):
 
 
 @dataclass
-class ChangePoint(_ChangePoint[TTestStats]):
+class ChangePointHunter(ChangePointOtava[TTestStats]):
```
henrikingo, line 75: @Sowiks Did you have some plan what to do with these almost identical change point classes?

pyproject.toml
```diff
@@ -100,6 +100,9 @@ filterwarnings = [
     "ignore::DeprecationWarning:future",
     "ignore::pytest.PytestCollectionWarning",
 ]
+console_output_style = "progress"
+junit_duration_report = "call"
+
```
henrikingo, line 103: Please ignore

otava/series.py
Sowiks: I was planning to consolidate these two classes.

Gerrrr:
@henrikingo CI is failing here. Can you take a look?

henrikingo:
Converting to draft. We should merge the two classes. This PR can eventually then be closed. Leaving it as draft for now.

Gerrrr:
@henrikingo in https://github.com/apache/otava/pull/154#issuecomment-4505152950 you mentioned that this PR should be merged first. Is there anything in this work stream I need to review now? Asking just to make sure I am not stalling your progress with reviews. In general, feel free to ping me if I am dropping the ball on your PRs that are ready.

henrikingo:
No and sorry if it wasn't clear. This PR currently renames two classes to emphasize that they are separate. Based on discussion (in this thread) what we should do is to merge the two classes into one.

This is IMO not a hard dependency for #154 but would of course be nicer to do this first.

henrikingo:
Actual suggestion for refactor is now in #161 
Closing this

PR #156, opened 2026-05-12
Update uv lockfile for urllib3 security fixes

arpitjain099:
## Summary
- Update `uv.lock` to bump `urllib3` from `2.6.3` to `2.7.0`.
- Keep changes lockfile-only and scoped to dependency resolution.

## Why
Dependabot reports two open high-severity advisories affecting `urllib3 < 2.7.0` in this repo lockfile. This update moves the resolved version to the patched range.

## Validation
- `uv sync --all-extras --dev`
- `uv run pytest -q` (fails in this environment for reasons unrelated to this lockfile change: tests invoke `uv` as a shell command not on PATH in subprocess contexts, and Docker-based E2E tests hit local runtime timeouts)

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1484,11 +1484,11 @@ wheels = [
 
 [[package]]
 name = "urllib3"
-version = "2.6.3"
+version = "2.7.0"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


lhotari:
Thanks for the contribution! There's already #155 by dependabot so I'll close this PR.

PR #159, opened 2026-05-17
Add and clarify manual release process steps

henrikingo:
* Clarify and separate the two svn repos https://dist.apache.org/repos/dist/dev/incubator/otava https://dist.apache.org/repos/dist/release/incubator/otava

* Use test.pypi.org earlier

* Link to otava.apache.org/download for downloadsin the final announcement

### DIFF
--- a/docs/RELEASE.md
+++ b/docs/RELEASE.md
@@ -25,9 +25,14 @@
 ## Introduction
 
 The Otava community treats releases with great importance. They are a public face of the project and most users interact with the project only through the releases. Releases are signed off by the entire Otava community in a public vote.
+
 Each release is executed by a Release Manager, who is selected/proposed by the Otava PMC members. This document describes the process that the Release Manager follows to perform a release. Any changes to this process should be discussed and adopted on the [dev@otava.apache.org](mailto:dev@otava.apache.org) mailing list.
 
-Please remember, that the act of publishing software has both legal and policy significance. This guide complements the foundation-wide [Product Release Policy](https://www.apache.org/dev/release.html) and [Release Distribution Policy](https://www.apache.org/dev/release-distribution).
+Please remember, that the act of publishing software has both legal and policy significance. This guide complements the foundation-wide
+
+* [Product Release Policy](https://www.apache.org/dev/release.html),
+* [Release Distribution Policy](https://www.apache.org/dev/release-distribution),and
+* [Release Download Pages for Projects](https://infra.apache.org/release-download-pages.html)
 
 ## Overview
 
@@ -38,6 +43,8 @@ Apache Otava release consists of:
 * PyPI wheels published to [pypi.org](https://pypi.org/project/apache-otava/).
 * Docker images published to [Dockerhub](https://hub.docker.com/r/apache/otava).
 * Release tag on [GitHub](https://github.com/apache/otava/releases).
+* All of the above published on [https://otava.apache.org/docs/downloads](https://infra.apache.org/release-download-pages.html)
+* ...and on dev@otava.apache.org and general@incubator.org
 
 ### Phases of the release
 
@@ -66,6 +73,32 @@ In general, the community prefers to have a rotating set of 3-5 Release Managers
 
 Before your first release, you should perform one-time configuration steps. This will set up your security keys for signing the release and access to various release repositories.
 
+## Paths
+
+For clarity, we will use the following environment variables when referring to different paths and repositories. You are free to use any path you desire for the values of these repositories:
+
+```bash
+# This is the main development repository, that you cloned from github.com/apache/otava
+export OTAVA_GIT_REPO=$HOME/git/apache/otava
+
+# This is an ASF subversion repository where you can publish preview/release candidate packages before the vote
+# https://dist.apache.org/repos/dist/dev/incubator/otava/
+export OTAVA_SVN_DEV=$HOME/svn/apache/otava-dist-dev-repo
+
+# This is an ASF subversion repository where you publish the tar files, images, etc. after the vor
+# https://dist.apache.org/repos/dist/release/incubator/otava/
+export OTAVA_SVN_REL=$HOME/svn/apache/otava-dist-release-repo
+```
+
+While at it, make sure the paths exist.
+
+```bash
+mkdir -p dirname $OTAVA_GIT_REPO
+mkdir -p dirname $OTAVA_SVN_DEV
+mkdir -p dirname $OTAVA_SVN_REL
+```
+
+
 ### One-time configuration
 
 #### GPG Key
@@ -90,8 +123,12 @@ Here, the key ID is the 8-digit hex string in the pub line: 845E6689.
 Now, add your Apache GPG key to the Otava's KEYS file in the [release](https://dist.apache.org/repos/dist/release/incubator/otava/KEYS) repository at [dist.apache.org](https://dist.apache.org/repos/dist/). Follow the instructions listed at the top of these files. (Note: Only PMC members have write access to the release repository. If you end up getting 403 errors ask on the mailing list for assistance.) PMC members can refer to the following scripts to add your Apache GPG key to the KEYS in the release repository.
 
 ```bash
-svn co https://dist.apache.org/repos/dist/release/incubator/otava otava-dist-release-repo
-cd otava-dist-release-repo
+# Checkout both the dev and release repositories up front
+svn co https://dist.apache.org/repos/dist/dev/incubator/otava $OTAVA_SVN_DEV
+svn co https://dist.apache.org/repos/dist/release/incubator/otava $OTAVA_SVN_REL
+
+# For GPG key we always use the dist/release repo:
+cd $OTAVA_SVN_REL
 (gpg --list-sigs <YOUR_KEY_ID> && gpg --armor --export <YOUR_KEY_ID>) >> KEYS
 svn ci -m "[otava] Add <YOUR_NAME>'s public key"
 ```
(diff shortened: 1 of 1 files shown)


henrikingo:
Follow-up activities to the release:

* This PR: Several additions and clarifications to docs/RELEASE.md
* Unless someone objects, I will move the previous releases 0.6.0, 0.7.0 and 0.7.1 from https://dist.apache.org/repos/dist/dev/incubator/otava/ to https://dist.apache.org/repos/dist/release/incubator/otava/
* Website, one remaining fix: The "all releases" part of https://otava.apache.org/docs/download should point to URLs that start with archive.apache.org or in any case never to https://dist.apache.org/repos/dist/dev/incubator/otava/
* IMO docker builds should also be published somewhere before voting. Proposal beloiw.


The above should fix everything that came up during the voting and review on the Incubator list. But one last comment is that this process has way too many manual steps. It is almost 2 days of work (for a first timer anyway) and very error prone due to the complexity. So I would like to propose as follow-up work that we automate a large part of this in Github CI. Specifically, the creation of source tar, pypi tar and whl, and docker image should be running in CI anyway. If the CI detects that user pushed a tag that has the form `x.y.z-incubating-rcN` then the resulting artifacts will also be attached to the result of the workflow. 

The release managers job is then just to download and carefully test and review these files, sign them, upload to dist.apache.org/.../dev/ and start the voting.

docs/RELEASE.md
```diff
@@ -66,6 +73,32 @@ In general, the community prefers to have a rotating set of 3-5 Release Managers
...
+
+```bash
+# This is the main development repository, that you cloned from github.com/apache/otava
+export OTAVA_GIT_REPO=$HOME/git/apache/otava
+
+# This is an ASF subversion repository where you can publish preview/release candidate packages before the vote
+# https://dist.apache.org/repos/dist/dev/incubator/otava/
+export OTAVA_SVN_DEV=$HOME/svn/apache/otava-dist-dev-repo
+
+# This is an ASF subversion repository where you publish the tar files, images, etc. after the vor
```
justinmclean: after the vote?

docs/RELEASE.md
```diff
@@ -90,8 +123,12 @@ Here, the key ID is the 8-digit hex string in the pub line: 845E6689.
...
-cd otava-dist-release-repo
+# Checkout both the dev and release repositories up front
+svn co https://dist.apache.org/repos/dist/dev/incubator/otava $OTAVA_SVN_DEV
+svn co https://dist.apache.org/repos/dist/release/incubator/otava $OTAVA_SVN_REL
+
+# For GPG key we always use the dist/release repo:
+cd $OTAVA_SVN_REL
 (gpg --list-sigs <YOUR_KEY_ID> && gpg --armor --export <YOUR_KEY_ID>) >> KEYS
 svn ci -m "[otava] Add <YOUR_NAME>'s public key"
 ```
```
justinmclean, line 134: this only needs to be done once per release manager

docs/RELEASE.md
```diff
@@ -220,12 +289,28 @@ sha512sum --check apache_otava-$RELEASE_VERSION-py3-none-any.whl.sha512
...
+```bash
+# Go to dev/preview svn directory
+cd $OTAVA_SVN_DEV
+cd  $RELEASE_VERSION-incubating-$RELEASE_CANDIDATE/pypi
+# Publish to Test PyPI
+twine upload --verbose --repository testpypi apache_otava-$RELEASE_VERSION-py3-none-any.whl apache_otava-$RELEASE_VERSION.tar.gz
+# Verify at https://test.pypi.org/project/apache-otava/
 ```
 
+
```
justinmclean, line 315: This must be clearly marked as a release candidate

docs/RELEASE.md
```diff
@@ -376,9 +464,12 @@ Once the release candidate has been reviewed and approved by the project and Inc
 In `otava-dist-release-repo`:
 
 ```bash
-cp -r $RELEASE_VERSION-incubating-$RELEASE_CANDIDATE $RELEASE_VERSION-incubating
+cp -r $OTAVA_SVN_DEV/$RELEASE_VERSION-incubating-$RELEASE_CANDIDATE $OTAVA_SVN_REL/$RELEASE_VERSION-incubating
+cd $OTAVA_SVN_REL
+svn update && svn status
```
justinmclean: It's probably better to use an SVN move command to transfer the RC from the dive to the dist area. If you do this, there's no need to check it out, and you're 100% that the RC you have voted on is the release.

henrikingo: Thanks. I wasn't sure if these are the same repo or independent/ if this is possible. Certainly makes sense if it is.

docs/RELEASE.md
henrikingo: This is a good point worth discussing.

I moved this earlier in the process, because:
* test.pypi.org in itself is 100% for testing purposes, and nothing published there can be considered a release. (Whether GA, RC or something else)
* It is however the case, that this step loses the "-rcN" provenance completely. The files uploaded are identical to the files that are later published as GA.
* or... if the vote fails, the files are not identical in their content but filenames and version strings are already same as they will be for the successful GA release, once it happens.
  * One reason to do it this way is that the sha512 signatures reference the file name of what they signed, so we cannot rename files (from *-rc1.tar.gz to *.tar.gz) without having to also edit or recreate contents of a file.
  * Essentially the model we are close to here is "channels", which I've mostly seen used in K8s ecosystem. Te binary or tar archve is built once, then promoted to different channels as it is condsidered

Now, if I interpret you correctly, we should rather adhere to a policy where release files may (in practice) only leave ASF maintained infrastructure once the vote has passed. This is possible to do (the previous 0.8.0-incubating release was done like that). The challenge in this approach is that there's a heightened risk we approve a release, then realize it can't be published on pypi or dockerhub due to some restriction we aren't meeting correctly.. Leading to a spiral of repeat votes  to get one release properly into teir channels.

***
So to answer your question/requirement succinctly: These files wold NOT be marked as -rcN, or anything else that identifies them as a release canidate, rather their context should presumably make it clear they are only publised for tesing.

docs/RELEASE.md
dave2wave: While it may be better to use SVN move. SVN import is also effective.

On another note. The new Apache Trusted Release platform is about to be ready for Alpha3 and this will automate many of the release steps.

docs/RELEASE.md
```diff
@@ -25,9 +25,14 @@
...
 
 The Otava community treats releases with great importance. They are a public face of the project and most users interact with the project only through the releases. Releases are signed off by the entire Otava community in a public vote.
+
 Each release is executed by a Release Manager, who is selected/proposed by the Otava PMC members. This document describes the process that the Release Manager follows to perform a release. Any changes to this process should be discussed and adopted on the [dev@otava.apache.org](mailto:dev@otava.apache.org) mailing list.
 
-Please remember, that the act of publishing software has both legal and policy significance. This guide complements the foundation-wide [Product Release Policy](https://www.apache.org/dev/release.html) and [Release Distribution Policy](https://www.apache.org/dev/release-distribution).
+Please remember, that the act of publishing software has both legal and policy significance. This guide complements the foundation-wide
+
+* [Product Release Policy](https://www.apache.org/dev/release.html),
+* [Release Distribution Policy](https://www.apache.org/dev/release-distribution),and
```
dave2wave: The url you give redirects to https://infra.apache.org/release-distribution.html which is best to use.

docs/RELEASE.md
dave2wave: Do not remove keys that have been used to sign releases even if those releases have been archived and the key is expired.

docs/RELEASE.md
dave2wave: @henrikingo - I agree for PyPi. With other distribution channels there is no real test and @justinmclean has a valid concern about downstream confusion.

docs/RELEASE.md
```diff
@@ -241,8 +326,10 @@ Hello everyone,
 
 Please review and vote for the releasing Apache Otava (incubating) $RELEASE_VERSION-$RELEASE_CANDIDATE.
 
-Changelog for this release candidate <>.
-The official Apache source release has been deployed to https://dist.apache.org/repos/dist/dev/incubator/otava/$RELEASE_VERSION-incubating-$RELEASE_CANDIDATE/.
+Release notes for this release candidate
+https://github.com/apache/otava/blob/$RELEASE_VERSION-incubating-$RELEASE_CANDIDATE/docs/release-notes/$RELEASE_VERSION-incubating
+
+The official ASF source release has been deployed to https://dist.apache.org/repos/dist/dev/incubator/otava/$RELEASE_VERSION-incubating-$RELEASE_CANDIDATE/.
```
dave2wave, line 334: No, it's https://dist.apache.org/repos/dist/release/incubator/otava/$RELEASE_VERSION-incubating-$RELEASE_CANDIDATE/.

If this is an external list then `https://dlcdn.apache.org/incubator/otava/` for the artifacts and `https://download.apache.org/incubator/otava/` for the signatures and checksums.

For archived release links - `https://archive.apache.org/dist/incubator/otava/`

Please remove old releases from https://dist.apache.org/repos/dist/dev/incubator/otava/ to help reduce the Terrabyte footprint of SVN Dist.

dave2wave (dismissed):

docs/RELEASE.md
henrikingo: > On another note. The new Apache Trusted Release platform is about to be ready for Alpha3 and this will automate many of the release steps.

Glad to hear. I would have filed a ticket to automate most of this otherwise. Happy to wait for a centralized solution.

docs/RELEASE.md
henrikingo: Dave, note that the email template you are commenting on here is the pre-vote email / the one to start the vote. Changing this is the main reason for this PR. The discussion on the general@incubator list was here: https://lists.apache.org/thread/rg8rkxmcp1llzcw10vpch28r66kw865s

docs/RELEASE.md
dave2wave: Next Otava release let's use ATR.

docs/RELEASE.md
dave2wave: Sorry I missed the detail about which email template was being discussed.

dave2wave (approved):

PR #161, opened 2026-05-31
Refactor: Unify ChangePoint and SignificanceTester

henrikingo:
Okay I don't know what to say... Originally I was just going to merge ChangePoint_ and ChangePoint and add the missing .metrics attribute to the survivor. Then it somehow escalated from there.

This needs much more test coverage, but sharing so you can see where I was going with this and also to get feedback on whether this makes the code better or worse...

ChangePoint:

    Unify the two ChangePoint classes and add container classes
    
    * Unify the ChangePoint_ class in hunter code and the new
      ChangePoint introduced by the new edivisive implementation
    
    Then it got out of hand a bit ...
    
    * Separate index and timestamp into different domains.
      cp.index is used in the context of a single metric and its
      history of results. Time and commit otoh are on the
      ChangePointGroup level (essentially a "row"). Note
      that different metrics can now have different cp.index
      for the same cpg.time or cpg.attributes['commit'], if
      they have a different history.
    
    * Introduce a ChangePoints class which is just a list of
      ChangePointGroups but actually comes with 2 different
      implementations.
    
    The last one is supposed to become the class you are left holding
    once all the change points are computed. Until now we had lots
    of nice classes for each step of computation, but in the end
    you were left holding a dict[str, ChangePointGroup]. The new
    class now encapsulates that dict,

SignificanceTester:

    This moves stats and functionality up towards parent classes
    so that generic stats like mean are always computed for
    all variants. In fact TTestStat is now an empty class,
    it's functionality fully absorbed by the parent. (But note
    that the class name/type itself carries information about the pvalue.


otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+    def to_json(self, rounded=True):
+        cps = ChangePointSerializer(self)
+        return cps.to_json(rounded)
+
+
+class ChangePointSerializer(ChangePoint):
+    """
+    Utility class with getters and json serialization for a ChangePoint.
+
+    TODO: Maintaining this is tedious. We should replace it with pydantic or some
```
Gerrrr, line 213: yes please!

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+        return iter([v for v in list(self.changes.values())])
+
+
+class ChangePoints:
+    """
+    A list of ChangePointGroup objects.
+
+    Typical usage of this would be to hold all the change points over a history of a single test,
+    the test producing one or more metrics. Note that this is a sparse structure: It is NOT
+    guaranteed that each row (each GhangePointGroup) has each metric. Similarly it is not guaranteed
```
Gerrrr: Could you elaborate and/or provide an example here of why this is a sparse structure?

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+    that a given metric will hold the full sequence.
+
+    Companion class ChangePointsByMetric is expected to provide functionally equivalent interface, but
+    storing each series separately by metric, which is used in parts of the code base, in particular, what
+    Series.analyze() returns.
+    Subclass ChangePointsByTime is this same class, but can be used if you explicitly want to mark the ordering.
+    """
+
+    def __init__(self, cps=None):
+        if isinstance(cps, dict) and not isinstance(cps, OrderedDict):
```
Gerrrr: This type signature is complex enough to deserve both `:param` in the block comment and a type signature.

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+        else:
+            # TODO: logging
+            # print(self.change_points)
+            # print(cpg)
+            raise ValueError(
+                "ChangePoints.append() can only be used such that time is monotonically increasing"
+            )
+
+    def extend(self, cps):
+        errmsg = "ChangePoints.extend() takes as argument a list of ChangePointGroup objects."
```
Gerrrr, line 447: This deserves a type signature and a comment with `:param` description.

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+            cps = [cps]
+        for obj in sorted(cps, key=lambda cpg: cpg.time):
+            if not isinstance(obj, ChangePointGroup):
+                t = type(obj)
+                raise TypeError(
+                    f"ChangePoints() takes as argument one or more ChangePointGroup objects. Got {t}."
+                )
+        self.change_points = cps
+
+    def append(self, cpg: ChangePointGroup):
```
Gerrrr, line 427: This deserves a type signature and a comment with `:param` description.

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+        """
+        by_metric = ChangePointsByMetric()
+
+        for row in sorted(self.change_points, key=lambda cpg: cpg.time):
+            assert isinstance(row, ChangePointGroup)
+            # append() does the necessary shuffling into separate columns
+            by_metric.append(row)
+
+
+class ChangePointsByTime(ChangePoints):
```
Gerrrr, line 540: Please add block comments for both classes motivating when one should use one or another.

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+    def pivot(self):
+        """
+        Return the same object as ChangePointsByMetric.
+        """
+        by_metric = ChangePointsByMetric()
+
+        for row in sorted(self.change_points, key=lambda cpg: cpg.time):
+            assert isinstance(row, ChangePointGroup)
+            # append() does the necessary shuffling into separate columns
+            by_metric.append(row)
```
Gerrrr, line 536: This method should return `by_metric`.

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+    def __len__(self):
+        return len(self.change_points)
+
+    def __getitem__(self, n):
+        return self.change_points[n]
+
+    def metrics(self) -> set:
+        all_metrics = set()
+        for row in self.change_points:
+            all_metrics.add(row.metrics())
```
Gerrrr: This throws a type error:

```python
cps = ChangePointsByTime()
cps.append(ChangePointGroup(time=1.0, attributes={}, changes={'m1': cp}))
cps.metrics()
# TypeError: unhashable type: 'dict_keys'
```

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+        cp_by_time = ChangePointsByTime()
+        for cpg in sorted(intermediate, key=lambda cpg: cpg.time):
+            cp_by_time.append(cpg)
+        return cp_by_time
+
+    def append(self, cpg: ChangePointGroup):
+        if not isinstance(cpg, ChangePointGroup):
+            raise TypeError("ChangePoints.append() takes as argument one ChangePointGroup.")
+        for metric in cpg.metrics():
+            self.change_points[metric].append(cpg.select_metrics(metric))
```
Gerrrr: `self.change_points[metric]` will throw if this is the first time we see this metric.

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+    """
+
+    def __init__(self, cps=None):
+        if isinstance(cps, ChangePointsByMetric):
+            self.change_points = cps.change_points
+        if isinstance(cps, ChangePointsByTime):
+            self.change_points = cps.pivot().change_points
+
+        if cps is None:
+            self.change_points = OrderedDict()
```
Gerrrr: `self.change_points` can be one of 3 types - `ChangePointsByMetric`, `ChangePointsByTime`, `OrderedDict`. OrderedDict is an outlier here as it does not share base class / interface with the other two, so it is harder to reason about and does require more explicit testing.

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+            for k, v in cps.items():
+                if not isinstance(k, float):
+                    raise TypeError(
+                        "ChangePointsByTime with OrderedDict() as constructor input requires the keys to be float (timestamps)?"
+                    )
+                if not isinstance(v, ChangePointGroup):
+                    raise TypeError(
+                        "ChangePointsByTime input must be an OrderedDict() of ChangePointGroup objects as values."
+                    )
+            self.change_points = sorted(cps, key=lambda cpg: cpg.time)
```
Gerrrr: This seems to call `.time` on the key (float), not the value.

```python
>>> from collections import OrderedDict
>>> from otava.change_point_divisive.base import (
...       BaseStats, ChangePoint, ChangePointGroup, ChangePointsByTime,
...   )
>>> cp = ChangePoint(index=1, qhat=0.0, stats=BaseStats([1, 1], [2, 2]), metric="s1")
>>> cpg = ChangePointGroup(time=1.0, attributes={}, changes={"s1": cp})
>>> od = OrderedDict([(1.0, cpg)])
>>> ChangePointsByTime(od)
Traceback (most recent call last):
  File "<python-input-20>", line 1, in <module>
    ChangePointsByTime(od)
    ~~~~~~~~~~~~~~~~~~^^^^
  File "/Users/asorokoumov/Projects/otava/otava/change_point_divisive/base.py", line 343, in __init__
    self.change_points = sorted(cps, key=lambda cpg: cpg.time)
                         ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/asorokoumov/Projects/otava/otava/change_point_divisive/base.py", line 343, in <lambda>
    self.change_points = sorted(cps, key=lambda cpg: cpg.time)
                                                     ^^^^^^^^
AttributeError: 'float' object has no attribute 'time'
```

otava/change_point_divisive/base.py
Gerrrr: Even better - replace the constructor with static builder methods like ` ChangePoints.from_dict`, `ChangePoints.from_list`, etc

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+                        )
+                self.change_points[metric] = sorted(cpglist, key=lambda cpg: cpg.time)
+
+    def pivot(self):
+        # Now we pivot (metric,time) to (time,metric) so that we return ChangePoints() objects
+        intermediate = []
+        for metric, points in self.change_points.items():
+            for cpg in sorted(points, key=lambda cpg: cpg.time):
+                assert isinstance(cpg, ChangePointGroup)
+                intermediate.append(cpg)
```
Gerrrr, line 609: You probably want to copy `cpg` here and below; otherwise we are mutating objects here:

```python
from otava.change_point_divisive.base import (
      BaseStats, ChangePoint, ChangePointGroup, ChangePointsByMetric,
  )

cp1 = ChangePoint(index=6, qhat=0.0, stats=BaseStats([1, 1], [2, 2]), metric="s1")
cp2 = ChangePoint(index=6, qhat=0.0, stats=BaseStats([1, 1], [2, 2]), metric="s2")

cpg_a = ChangePointGroup(time=6.0, attributes={}, changes={"s1": cp1})
cpg_b = ChangePointGroup(time=6.0, attributes={}, changes={"s2": cp2})

by_metric = ChangePointsByMetric({"s1": [cpg_a], "s2": [cpg_b]})

print("before:", cpg_a.changes)  # {'s1': cp1}
by_metric.pivot()
print("after 1st pivot:", cpg_a.changes)  # {'s1': cp1, 's2': cp2}, mutated
by_metric.pivot()  # KeyError: "Duplicate keys. Shouldn't happen."
```

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+                continue
+            for metric in cpg.metrics():
+                cpg[metric] = c
+
+        return cpg
+
+    def metrics(self):
+        return set(self.change_points.keys())
+
+    def select_metrics(self, m: list[str] | str):
```
Gerrrr, line 688: This method is little tricky. By selecting a non-existing metric we can get a full list of empty changepoints.

otava/change_point_divisive/base.py
```diff
@@ -14,61 +14,640 @@
...
+        return filtered
+
+    def set(self, metric: str, cp: ChangePoint):
+        self.changes[metric] = cp
+
+    def __iter__(self):
+        return iter([v for v in list(self.changes.values())])
+
+
+class ChangePoints:
```
Gerrrr, line 341: Is this supposed to be the internal abstract class while `ChangePointsByTime` and `ChangePointsByMetric` are "public" classes?

Gerrrr (commented):
This refactoring makes sense to me directionally. 

As the main feedback at this stage, I would prefer these data structures to be more rigid type-wise OR, even better, structured in a way that the reader understand the interface of arguments/parameters. Verbose type signatures everywhere would help as well.

Maybe 8 years of Java development destroyed by sense of beauty, but I found it hard to reason about behavior here primarily due to type polymorphism.

otava/change_point_divisive/base.py
henrikingo: Because at some point in time user may have added or discontinued tracking of some metrics, or indeed  entire tests. Or some metric just failed to register. This is okay and should be supported. (I didn't test whether it actually works yet, could require more work elsewhere.)

But you probably meant I should add to the comment?

otava/change_point_divisive/base.py
henrikingo: Initially I intended this to be the public class and by-time and by-metric would simply be internally different implementations, but identical in every way except performance.

I guess they are still identical APIs if you consider the methods they provide. If you access ChangePoitnts.change_points directly, then it will indeed have different structure. But maybe you are not supposed to do that?

otava/change_point_divisive/base.py
henrikingo: > Even better - replace the constructor with static builder methods like ` ChangePoints.from_dict`, `ChangePoints.from_list`, etc

Yeah, should have done this all along. Thanks, I'll break the constructor apart to such methods.

otava/change_point_divisive/base.py
henrikingo: > self.change_points can be one of 3 types - ChangePointsByMetric, ChangePointsByTime, OrderedDict

I don't follow? How could ChangePointsByMetric.change_points itself be of type ChangePointsByMetric?

otava/change_point_divisive/base.py
henrikingo: if self.change_points[] doesn't have "metric" then how would it ever get past line 570?

otava/change_point_divisive/base.py
Gerrrr: Thanks for the clarification! Yeah, a comment in the code would help.

henrikingo:
@Gerrrr What is wrong with the test failure here? It's unclear to me what exactly is failing?

otava/series.py
```diff
@@ -445,12 +365,12 @@ def metric(self, name: str) -> Metric:
 
     def to_json(self):
         change_points_json = {}
-        for metric, cps in self.change_points.items():
-            change_points_json[metric] = [cp.to_json(rounded=False) for cp in cps]
+        for cps in self.change_points:
+            change_points_json = [cp.to_json(rounded=False) for cp in cps]
```
Gerrrr: Before this change, we were adding adding a new `metric` -> `list-of-changepoints-as-json`. With your change, we override an empty dict with `list-of-changepoints-as-json`.

otava/series.py
```diff
@@ -445,12 +365,12 @@ def metric(self, name: str) -> Metric:
...
 
         weak_change_points_json = {}
-        for metric, cps in self.weak_change_points.items():
-            weak_change_points_json[metric] = [cp.to_json(rounded=False) for cp in cps]
+        for cps in self.weak_change_points:
```
Gerrrr: same as above

otava/change_point_divisive/base.py
```diff
@@ -40,35 +153,500 @@ class BaseStats:
...
+        }
+
+    def __getitem__(self, metric):
+        return self.changes[metric]
+
+    def metrics(self):
+        return self.changes.keys()
+
+    def commit(self, idx: int):
+        return self.attribute_at(idx).get("commit")
```
Gerrrr: There is no `attribute_at`.

otava/change_point_divisive/base.py
Gerrrr: as a nit, I'd make `change_points` "private" by renaming it into `_change_points` per Python's convention.

otava/change_point_divisive/base.py
Gerrrr: ah, my bad - we do extract `cp.change_points` instead of assigning `self.change_points = cp`.

otava/change_point_divisive/base.py
Gerrrr: My bad, I meant `ChangePointsByTime`:

```python
>>> from otava.change_point_divisive.base import (
...       BaseStats, ChangePoint, ChangePointGroup,
...       ChangePointsByTime, ChangePointsByMetric,
...   )
>>> cp = ChangePoint(index=1, qhat=0.0, stats=BaseStats([1, 1], [2, 2]), metric="m1")
>>> cpg = ChangePointGroup(time=1.0, attributes={}, changes={"m1": cp})
>>> by_time = ChangePointsByTime(); by_time.append(cpg)
>>> len(by_time.select_metrics("typo").change_points)
>> by_time.select_metrics("typo").change_points[0]
ChangePointGroup(time=1.0, attributes={}, changes={})
```

Gerrrr: Also, the errors are inconsistent where they are present. For example:

```python
>> by_time.get_change_points_for_metric("typo")
Traceback (most recent call last):
  File "<python-input-8>", line 1, in <module>
    by_time.get_change_points_for_metric("typo")
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/asorokoumov/Projects/otava/otava/change_point_divisive/base.py", line 436, in get_change_points_for_metric
    return [list(cpg.changes.values())[0] for cpg in single_metric.change_points]
            ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
```

but:

```python
ChangePointsByMetric({"m1": [cpg]}).select_metrics("typo")
Traceback (most recent call last):
  File "<python-input-9>", line 1, in <module>
    ChangePointsByMetric({"m1": [cpg]}).select_metrics("typo")
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/asorokoumov/Projects/otava/otava/change_point_divisive/base.py", line 571, in select_metrics
    filtered.change_points[metric] = self.change_points[metric]
                                     ~~~~~~~~~~~~~~~~~~^^^^^^^^
KeyError: 'typo'
```

We should extract invariant-checking methods into the base class and invoke those on each operation.

otava/change_point_divisive/base.py
```diff
@@ -40,35 +153,500 @@ class BaseStats:
...
+    def items(self):
+        return self.change_points.items()
+
+    def __iter__(self):
+        return self.pivot().__iter__()
+
+    def __len__(self):
+        return max([len(cpg) for metric, cpg in self.change_points.items()])
+
+    def __getitem__(self, n):
```
Gerrrr, line 670:
```python
>> from otava.change_point_divisive.base import (
...       BaseStats, ChangePoint, ChangePointGroup, ChangePointsByMetric,
...   )
>>> cp_a = ChangePoint(index=1, qhat=0.0, stats=BaseStats([1, 1], [2, 2]), metric="m1")
>>> cp_b = ChangePoint(index=2, qhat=0.0, stats=BaseStats([3, 3], [4, 4]), metric="m2")
>>> by_metric = ChangePointsByMetric({
...       "m1": [ChangePointGroup(time=1.0, attributes={}, changes={"m1": cp_a})],
...       "m2": [ChangePointGroup(time=2.0, attributes={}, changes={"m2": cp_b})],
...   })
>>> by_metric[0]
...   # TypeError: 'ChangePointGroup' object does not support item assignment
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    by_metric[0]
    ~~~~~~~~~^^^
  File "/Users/asorokoumov/Projects/otava/otava/change_point_divisive/base.py", line 553, in __getitem__
    cpg[metric] = c
    ~~~^^^^^^^^
TypeError: 'ChangePointGroup' object does not support item assignment
```

Gerrrr: This kind of errors in Python makes me want to type-check all method arguments and intermediate variables.

otava/change_point_divisive/base.py
henrikingo: In the new push I did copy() as a deep copy but pivot() is just another view of the same data.

Yo can always do copy().pivot() if desired.

otava/change_point_divisive/base.py
henrikingo: Reading this myself now... This is written by claude and it makes no sense. I'll have to simplify. I have to re-read what else it might ave edone.

PS. Sorry we seem to have criss crossed today. Most of your comments are still useful. But I'll make a top level comment soon when I'm done with  this batch.

Gerrrr: 👍  please ping me when you need another review.

otava/change_point_divisive/base.py
henrikingo: values()[0] is bad code anyway. Once I straighten this out, there should also be the usual KeyErrors etc

henrikingo:
Okay so this should represent a bit more serious effort to create a complete pantheon of ChangePoints/ChangePointGroup/ChangePoint/Candidate/Stats classes and with better test coverage too.

On the todo side I still need to add the typing info to many functions, possibly some also need a docstring or :param: type info. Also I want to run pytest --coverage to ensure it's good.

henrikingo:
Okay @Gerrrr and others, I think this is the first attempt at something actually worth reviewing seriously. Hopefully we can merge this to unlock 3 other pr's waiting on this one.

I've done some docstrings at strategic locations, but in the end I guess I did fairly little :param: type of docs. Code should be easier to follow for example because constructors were broken into 3-4 different factory methods each of which does one thing only.

Happy to add more comments if needed, but also didn't want to overdo it and curious to see your reaction now.

On the testing side I installed pytest-cov to keep me honest, and then did something I never did before: 100% code covreage with unit tests! I feel like this could become a thing.

otava/report.py
```diff
@@ -74,19 +75,19 @@ def __format_log_annotated(self, test_name: str) -> str:
...
+        for cpg in self.__change_points:
             separator = ""
             info = ""
             for col_index, col_name in enumerate(columns):
                 col_width = col_widths[col_index]
-                change = [c for c in cp.changes if c.metric == col_name]
+                change = [c for m, c in cpg.changes.items() if m == col_name]
                 if change:
-                    change = change[0]
+                    change = ChangePointSerializer(change[0])
```
Gerrrr, line 90:
```python
>>> from otava.series import Metric, Series
... from otava.report import Report, ReportType
... s = Series('t', None, list(range(11)),
...     {'a': Metric(1, 1.0), 'b': Metric(1, 1.0)},
...     {'a': [1.02,0.95,0.99,1.00,1.12,0.90,0.50,0.51,0.48,0.48,0.55],
...      'b': [2.02,2.03,2.01,2.04,1.82,1.85,1.79,1.81,1.80,1.76,1.78]},
...     {})
>>> Report(s, s.analyze().change_points_by_time).produce_report('t', ReportType.REGRESSIONS_ONLY)
...
Traceback (most recent call last):
  File "<python-input-0>", line 8, in <module>
    Report(s, s.analyze().change_points_by_time).produce_report('t', ReportType.REGRESSIONS_ONLY)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/asorokoumov/Projects/otava/otava/report.py", line 57, in produce_report
    return self.__format_regressions_only(test_name)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/Users/asorokoumov/Projects/otava/otava/report.py", line 113, in __format_regressions_only
    metric = self.__series.metrics[cp.metric]
                                   ^^^^^^^^^
AttributeError: 'str' object has no attribute 'metric'
```

`LOG` and `JSON` work fine.

otava/bigquery.py
```diff
@@ -87,9 +87,10 @@ def insert_change_point(
         test: BigQueryTestConfig,
         metric_name: str,
         attributes: Dict,
-        change_point: ChangePoint,
+        change_point_group: ChangePointGroup,
     ):
-        kwargs = {**attributes, **{test.time_column: datetime.utcfromtimestamp(change_point.time)}}
+        change_point = change_point_group[metric_name]
```
Gerrrr:
```python
>>> from otava.change_point_divisive.base import ChangePoint, ChangePointGroup, BaseStats
>>> stats = BaseStats(pvalue=0.001, mean_1=1.0, mean_2=2.0, std_1=0.0, std_2=0.0)
>>> cpg = ChangePointGroup(time=100.0, attributes={'commit':'sha'},
...                        changes={'m': ChangePoint(index=3, qhat=1.0, stats=stats, metric='m')})
...
>>> cpg['m'].forward_change_percent()
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    cpg['m'].forward_change_percent()
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ChangePoint' object has no attribute 'forward_change_percent'
```

It should probably be:

```python
change_point = ChangePointSerializer(change_point_group[metric_name])
```

like in `postgres.py`

otava/series.py
Gerrrr:
```python
>>> from otava.series import Metric, Series
>>> s = Series('t', None, list(range(11)),
...     {'a': Metric(1,1.0), 'b': Metric(1,1.0)},
...     {'a': [1.02,0.95,0.99,1.00,1.12,0.90,0.50,0.51,0.48,0.48,0.55],
...      'b': [2.02,2.03,2.01,2.04,1.82,1.85,1.79,1.81,1.80,1.76,1.78]},
...     {})
...
>>> j = s.analyze().to_json()
>>> print(type(j['change_points']))
<class 'list'>
>>> print(len(j['change_points']), j['change_points'][0]['metric'])
1 a
```

This is an incorrect answer because we lost `b`. At the very least, we should do `change_points_json[metric] = ` instead of change_points_json = `.

While at it, let's take a step back evaluate the following scenario 

Looking at `to_json`/`from_json` a bit more, I'd say we have a problem here. Say, we have a new entry time with `time=100, commit=abc123` and a changepoint that regressed 2 metrics - `latency` got +50% and throughput `-20%`. In memory, this is a single `ChangePointGroup` with 2 changepoints. 

If we apply the fix described above, it would look like:

```json
{
  "latency":    [{"metric": "latency",    "index": 5, "forward_change_percent": 50, ...}],
  "throughput": [{"metric": "throughput", "index": 5, "forward_change_percent": -20, ...}]
}
```
Note that `time` and `commit` attributes are missing. If we to write a roundtrip test (we should!), it would fail here.

Wouldn't it be cool if `to_json` persisted the attributes, so we'd get:

```json
{
  "time": 100.0,
  "attributes": {"commit": "abc123"},
  "changes": [
    {"metric": "latency",    "index": 5, "forward_change_percent": 50, ...},
    {"metric": "throughput", "index": 5, "forward_change_percent": -20, ...}
  ]
}
```

otava/change_point_divisive/significance_test.py
```diff
@@ -37,6 +37,17 @@ class PermutationStats(BaseStats):
     extreme_qhat_perm: int
     n_perm: int
 
+    def copy(self):
+        # replace() preserves the subclass; deep-copy the array so the copy is independent
```
Gerrrr: this comment is not true :)

```python
>> import numpy as np
>>> from otava.change_point_divisive.significance_test import PermutationStats
>>> qhats = np.array([0.1, 0.2, 0.3])
>>> s = PermutationStats(pvalue=0.01, mean_1=1, mean_2=2, std_1=0, std_2=0,
...                      permuted_qhats=qhats, extreme_qhat_perm=1, n_perm=3)
>>> c = s.copy()
...
>>> print(c.permuted_qhats is s.permuted_qhats)
True  # this should be have been false if that would be a deep-copy because 2 copies would have had different pointers.
>>> c.permuted_qhats[0] = -999.0
>>> print(s.permuted_qhats[0] == -999.0)
True
```

tests/report_test.py
```diff
@@ -40,6 +40,13 @@ def series():
 
 @pytest.fixture(scope="module")
 def change_points(series):
+    # o = AnalysisOptions()
```
Gerrrr: nit: let's remove commented WIP code before merging this PR

tests/series_test.py
```diff
@@ -90,7 +145,7 @@ def test_div_by_zero():
     cpjson = analyzed_series.to_json()
     assert cpjson
     assert len(change_points) == 2
-    assert change_points[0].index == 3
```
Gerrrr, line 93: Should `assert cpjson` assert on the content of that variable?

tests/change_point_classes_test.py
```diff
@@ -0,0 +1,773 @@
...
+
+    with pytest.raises(TypeError):
+        c.extend("xxx")
+
+    with pytest.raises(TypeError):
+        c.extend(["a", "b"])
+
+
+def test_cpbm_getitem():
+    _ = ChangePointsByMetric.from_list(
```
Gerrrr: This looks like an unfinished/WIP test. We should either finish or remove it.

tests/change_point_classes_test.py
```diff
@@ -0,0 +1,773 @@
...
+
+def test_bymetric_by_time_roundtrip():
+    by_time = ChangePointsByTime.from_list([make_group(1.0, "a"), make_group(2.0, "b")])
+    bm = by_time.pivot()
+    again = bm.by_time()
+    assert isinstance(again, ChangePointsByTime)
+    assert [g.time for g in again] == [1.0, 2.0]
+
+
+def test_by_time_by_time_self():
```
Gerrrr: This seems identical to a test on the line 520.

tests/change_point_classes_test.py
```diff
@@ -0,0 +1,773 @@
...
+    by_time = ChangePointsByTime.from_list([make_group(1.0, "a"), make_group(2.0, "b")])
+    same = by_time.by_time()
+    # Yes it's the same object, a no-op, not a copy. (Open to other opinions here)
+    assert by_time == same
+    assert by_time.at_timestamp(1.0) == same.at_timestamp(1.0)
+
+
+def test_by_time_bm():
```
Gerrrr, line 485: This test is identical to `test_justcp_bm`

tests/change_point_classes_test.py
```diff
@@ -0,0 +1,773 @@
...
+    by_time = ChangePointsByTime.from_list([make_group(1.0, "a"), make_group(2.0, "b")])
+    same = by_time.by_metric()
+    # Yes it's the same object, a no-op, not a copy. (Open to other opinions here)
+    assert by_time != same
+    assert by_time.at_timestamp(1.0) == same.at_timestamp(1.0)
+
```
Gerrrr, line 491: It would be cool to have test asserting semantical parity between `ByTime` and `ByMetric`. Maybe property-based tests?

otava/change_point_divisive/base.py
Gerrrr: Bump on this one. It would be awesome to get rid of it before merging this PR or at least create a follow-up issue.

otava/change_point_divisive/base.py
```diff
@@ -40,39 +155,619 @@ class BaseStats:
...
+        return max([len(cpg) for metric, cpg in self._change_points.items()])
+
+    def __getitem__(self, n):
+        """
+        Returns the same ChangePointGroup as (ChangePointsByTime)[n] would.
+
+        Hence this calls pivot() and is slow.
+        Alternatively please use .at_timestamp(), .at_commit() or .get_change_points_for_metric() instead.
+        """
+        return self.by_time()[n]
```
Gerrrr, line 677: Thinking out loud, we may want to cache both views. Python's `__get__` is expected to work at O(1). Here, we do full rebuild, then return a single item.

otava/series.py
```diff
@@ -216,14 +137,17 @@ class AnalyzedSeries:
 
     __series: Series
     options: AnalysisOptions
-    change_points: Dict[str, List[ChangePoint]]
-    change_points_by_time: List[ChangePointGroup]
-    change_points_timestamp: Any
+    change_points: ChangePointsByMetric
+    change_points_by_time: ChangePointsByTime
+    change_points_timestamp: ChangePointsByMetric
```
Gerrrr: Should this be `change_points_timestamp: ChangePointsByTime`?

otava/change_point_divisive/base.py
```diff
@@ -40,39 +155,619 @@ class BaseStats:
...
+        :return: A deep copy of self, recursively calls also stats.copy().
+        """
+        return ChangePoint(
+            index=self.index, qhat=self.qhat, stats=self.stats.copy(), metric=self.metric
+        )
 
     def __eq__(self, other):
-        '''Helpful to identify new Change Points during divisive algorithm'''
+        """Helpful to identify new Change Points during divisive algorithm"""
         return isinstance(other, self.__class__) and self.index == other.index
```
Gerrrr: I am not sure comparing change points by index only is correct:


```python
>> from otava.change_point_divisive.base import ChangePoint, BaseStats
>>> stats = BaseStats(pvalue=0.001, mean_1=1.0, mean_2=2.0, std_1=0.1, std_2=0.1)
>>> cp_latency    = ChangePoint(index=5, qhat=1.0, stats=stats, metric="latency")
>>> cp_throughput = ChangePoint(index=5, qhat=1.0, stats=stats, metric="throughput")
>>> print(cp_latency == cp_throughput)
True
>>> all_cps = [cp_latency, cp_throughput]
>>> all_cps.remove(ChangePoint(index=5, qhat=0.0, stats=stats, metric="throughput"))
>>> print(all_cps)
[ChangePoint(index=5, qhat=1.0, stats=BaseStats(pvalue=0.001, mean_1=1.0, mean_2=2.0, std_1=0.1, std_2=0.1), metric='throughput')]
```

Gerrrr (commented):
I think this refactoring is a huge improvement of the existing API and I am looking forward to its landing. 

I did find a few bugs and left a suggestion about `to_json/from_json`. In general, it would make sense to add a round-trip test for them.

otava/change_point_divisive/base.py
henrikingo: It most certainly isn't in the general case and even here one should at least require `other.metric == self.metric`

I poisoned this with an assert False and turns out there were only 2 places that used this. Changed them to explicitly check for .index and then removed this __eq__ method.

otava/change_point_divisive/base.py
henrikingo: I'll create a follow up issue, at least provided that fixing the bugs already happening here isn't more work than migrating.

otava/change_point_divisive/base.py
henrikingo: Yes, I've been thinking it would actually be straightforward to essentially change ChangePoints class so that it **always** creates and maintains an up to date copy of both ChangePointsByMetric and ChangePointsByTime. But I intend to leave that as a follow-up task too and here I will focus on just getting the API right. Note that what I'm doing here should not be worse than we already have, as AnalyzedSeries maintans two separate collections of this and it is possible (although unlikekly) that you create an object where one is computed (by_metrict) but the other one is not (by_time).

otava/bigquery.py
henrikingo: ...and the reason this is untested because bigquery is a cloud-only database so can't test it without internet?

otava/series.py
henrikingo: Actually it should be datetime. (See 5 lines below) Thanks!

otava/series.py
henrikingo: You're right about the ChangePointGroups not serializing correctly. The glass half full version of this story is that it was easy to fix with the current ChangePoints* class and hierarcy of data structures.

Clearly the intent has been to do something that follows the by_metric() structure, so I fixed it based on that.

As for the latter half of your comment, the serialization doesn't lose data, rather timestamps and attributes are part of the test results data (Series) . A change point only has the minimal info about the change itself, and reference back to the overall Series, such as metric and index.

henrikingo: punting on the test as I don't expect these custom to_json/from_json to survive the 0.8.x series...

tests/series_test.py
henrikingo: I think this particular test is testing something rather specific (given its name) so I won't add generic assertions here

henrikingo (commented):
Went through all comments in a single pass...

henrikingo:
@Gerrrr ready for another review round

otava/change_point_divisive/base.py
henrikingo: Btw, thought about this after sleeping on it...

I think these to_json/from_json methods are currently overloaded. For example the rounded=True functionality is added (by me) because someone (me) used these methods to produce json that is sent to an API and a user interface. But I realized when working on this patch, that should rather be done in the Report class (which I don't), and these methods should be limited to serialization and deserialization, such as persisting the data to a database. If we could separate and limit functionality like that, perhaps these functions could be simplified a lot and there wouldn't be any issue.

otava/change_point_divisive/base.py
```diff
@@ -40,39 +155,634 @@ class BaseStats:
...
+        return self._change_points.items()
+
+    def keys(self):
+        return self._change_points.keys()
+
+    def __iter__(self):
+        return self.pivot().__iter__()
+
+    def __len__(self):
+        return max([len(cpg) for metric, cpg in self._change_points.items()])
```
Gerrrr, line 668: This fails on an empty input:

```python
>>> from otava.series import Series, Metric
>>> series = Series(
...       "flat",
...       branch=None,
...       time=[0, 1, 2, 3],
...       metrics={"m": Metric()},
...       data={"m": [1, 1, 1, 1]},
...       attributes={},
...   )
>>> analyzed = series.analyze()
>>> len(analyzed.change_points)
Traceback (most recent call last):
  File "<python-input-19>", line 1, in <module>
    len(analyzed.change_points)
    ~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/asorokoumov/Projects/otava/otava/change_point_divisive/base.py", line 666, in __len__
    return max([len(cpg) for metric, cpg in self._change_points.items()])
ValueError: max() iterable argument is empty
```

otava/series.py
```diff
@@ -386,32 +295,37 @@ def append(self, time, new_data, attributes):
...
                 weak_change_points[metric].append(
-                    ChangePoint(
-                        index=c.index, qhat=0.0, time=self.__series.time[c.index], metric=metric, stats=c.stats
+                    ChangePointGroup(
+                        time=self.__series.time[c.index],
+                        changes={metric: c},
+                        attributes=self.__series.attributes_at(c.index),
                     )
                 )
+            # TODO: Remove this. It should not be a requirement that metrics have the same history.
```
Gerrrr, line 330: What about this TODO?

otava/series.py
```diff
@@ -386,32 +295,37 @@ def append(self, time, new_data, attributes):
...
-        # If some metrics didn't participate in this round, we still keep them, but update the ones
-        # We did recompute
-        for metric in result.keys():
-            self.change_points[metric] = result[metric]
-        for metric in weak_change_points.keys():
-            self.weak_change_points[metric] = weak_change_points[metric]
-        self.change_points_by_time = self.__group_change_points_by_time(self.__series, self.change_points)
-        return result, weak_change_points
+        r = ChangePointsByMetric.from_dict(result)
+        w = ChangePointsByMetric.from_dict(weak_change_points)
```
Gerrrr, line 334: It looks like we also need to update ChangePointsByTime:

```python
>> from otava.series import Series, Metric
>>> series1 = [1.02, 0.95, 0.99, 1.00, 1.12, 0.90, 0.50, 0.51, 0.48, 0.48, 0.55]
>>> series2 = [2.02, 2.03, 2.01, 2.04, 1.82, 1.85, 1.79, 1.81, 1.80, 1.76, 1.78]
>>> series = Series(
...       "test",
...       branch=None,
...       time=list(range(len(series1))),
...       metrics={"series1": Metric(), "series2": Metric()},
...       data={"series1": series1, "series2": series2},
...       attributes={},
...   )
>>>
>>> analyzed = series.analyze()
>>> analyzed.append(
...       time=[11, 12],
...       new_data={"series2": [33.33, 46.46]},
...       attributes={},
... )
(<otava.change_point_divisive.base.ChangePointsByMetric object at 0x104cb2ea0>, <otava.change_point_divisive.base.ChangePointsByMetric object at 0x1129556d0>)
>>> by_metric = [
...       int(cp.index)
...       for cp in analyzed.change_points.get_change_points_for_metric("series2")
... ]
>>> by_time = [
...       (group.time, sorted(group.changes))
...       for group in analyzed.change_points_by_time
... ]
>>> print("by_metric:", by_metric)
by_metric: [4, 11]
>>> print("by_time:", by_time)
by_time: [(4, ['series2']), (6, ['series1'])]
```

Gerrrr, line 334: We build and return `w`, but we never assign it. Is this correct?

otava/bigquery.py
Gerrrr: > the reason this is untested because bigquery is a cloud-only database so can't test it without internet

Yes :(

otava/series.py
Gerrrr: Thanks for fixing the immediate bug! 

We still have issues like:

```python
>>> from otava.series import Metric, Series, AnalyzedSeries
>>> s  = Series(
...       "t",
...       None,
...       list(range(11)),
...       {"a": Metric(1, 1.0), "b": Metric(1, 1.0)},
...       {
...           "a": [1.02,0.95,0.99,1.00,1.12,0.90,0.50,0.51,0.48,0.48,0.55],
...           "b": [2.02,2.03,2.01,2.04,1.82,1.85,1.79,1.81,1.80,1.76,1.78],
...       },
...       {},
...   )
>>> j = s.analyze().to_json()
>>> AnalyzedSeries.from_json(j)
Traceback (most recent call last):
  File "<python-input-8>", line 1, in <module>
    AnalyzedSeries.from_json(j)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/Users/asorokoumov/Projects/otava/otava/series.py", line 419, in from_json
    mean_1=cp["mean_before"],
           ~~^^^^^^^^^^^^^^^
KeyError: 'mean_before'
```

but I don't mind addressing them in a follow-up.

otava/series.py
```diff
@@ -445,12 +359,18 @@ def metric(self, name: str) -> Metric:
...
+        cpbm = self.change_points.by_metric()
+        for metric_name in self.change_points.metrics():
+            change_points_json[metric_name] = []
+            for cp in cpbm.select_metrics(metric_name):
+                change_points_json[metric_name].append(cp.to_json(rounded=False))
 
         weak_change_points_json = {}
-        for metric, cps in self.weak_change_points.items():
-            weak_change_points_json[metric] = [cp.to_json(rounded=False) for cp in cps]
+        wcpbm = self.change_points.by_metric()
```
Gerrrr: Should this use `self.weak_change_points`?

otava/series.py
```diff
@@ -503,9 +423,7 @@ def from_json(cls, analyzed_json):
                     pvalue=cp["pvalue"],
                 )
                 new_list.append(
-                    ChangePoint(
-                        index=cp["index"], time=cp["time"], metric=cp["metric"], stats=stat
-                    )
+                    ChangePoint(index=cp["index"], time=cp["time"], metric=cp["metric"], stats=stat)
```
Gerrrr: > time=cp["time"]

`ChangePoint` no longer accepts `time`, right?

Gerrrr (commented):
I think we are almost there! Also, apologies for long review iterations.

Gerrrr:
How is it going @henrikingo ? This PR is so close to finally getting merged!

otava/change_point_divisive/base.py
Gerrrr: 90eb9d1

otava/series.py
Gerrrr: e087bb0

Gerrrr: 62ce90f

otava/series.py
Gerrrr: c2782b8

otava/series.py
Gerrrr: 366fe6e

otava/series.py
Gerrrr: 366fe6e

PR #162, opened 2026-06-02
Loosen numpy constraint from ==2.2.0.* to ==2.2.*

jordancharest:
## What

Loosen the numpy version constraint for Python < 3.14 from `==2.2.0.*` to `==2.2.*`.

## Why

The inline comment in `pyproject.toml` already documents the intent as:

> NumPy 2.2.x supports Python 3.10–3.13 [...]

However, `==2.2.0.*` is a prefix match that pins to exactly `2.2.0` (numpy publishes no 4-segment versions), so it excludes the `2.2.1`–`2.2.6` patch releases along with their bug/security fixes. This looks unintentional: the dependency was a minor-level pin (`numpy==1.24.*`) until #108 bumped it to `2.2.0.*`, where the natural equivalent would have been `2.2.*`.

Loosening to `==2.2.*` makes the constraint match the documented intent and lets users pick up 2.2.x patch releases, while still excluding 2.3 (which is handled separately for Python ≥ 3.14).

## Changes

- `pyproject.toml`: `numpy==2.2.0.*` → `numpy==2.2.*` (Python < 3.14)
- `uv.lock`: regenerated via `uv lock` (numpy `2.2.0` → `2.2.6`)

## Testing

- `uv run pytest` (tests + perf): **112 passed** against numpy 2.2.6
- `uv run tox -e lint`: passed (ruff, autoflake, isort, flake8, license headers)


Gerrrr (commented):
Thanks @jordancharest !

PR #166, opened 2026-08-04
Make CSV importer configurable via ConfigArgParse

### LINKED ISSUE #119: Make CSV importer configurable via ConfigArgParse
CSV is the only importer that is not configurable via ConfigArgParse - env variables and CLI. We should fix that.

MrlixiangWE:
Problem

CSV is the only importer whose delimiter and quote character cannot be set through ConfigArgParse. They can currently only be configured per test in YAML.

Before / after

Before: CSV delimiter and quote character can only be configured through per-test `csv_options` in YAML.

After: they can also be set through `--csv-delimiter` and `--csv-quote-char`, the matching environment variables, or a `csv` section in the config file. Explicit global values override per-test options; otherwise existing behavior is unchanged.

Verification

`uv run pytest -q --ignore=tests/graphite_e2e_test.py --ignore=tests/postgres_e2e_test.py`

`uv run tox -e lint`

`uv build --no-create-gitignore`

Closes #119

### DIFF
--- a/docs/CSV.md
+++ b/docs/CSV.md
@@ -34,8 +34,24 @@ tests:
     metrics: [metric1, metric2]
     csv_options:
       delimiter: ','
-      quotechar: "'"
+      quote_char: "'"
 ```
+
+## CSV options
+
+The delimiter and quote character can also be configured globally:
+
+```yaml
+csv:
+  delimiter: ';'
+  quote_char: "'"
+```
+
+The corresponding command-line options are `--csv-delimiter` and
+`--csv-quote-char`. They can also be set through the `CSV_DELIMITER` and
+`CSV_QUOTE_CHAR` environment variables. Explicit global values override the
+`csv_options` values of individual tests; without a global value, existing
+per-test settings and defaults are unchanged.
 
 ## Example
 
--- a/otava/config.py
+++ b/otava/config.py
@@ -23,6 +23,7 @@ import configargparse
 from ruamel.yaml import YAML
 
 from otava.bigquery import BigQueryConfig
+from otava.csv_options import CsvConfig
 from otava.grafana import GrafanaConfig
 from otava.graphite import GraphiteConfig
 from otava.influxdb import InfluxDBConfig
@@ -34,6 +35,7 @@ from otava.util import merge_dict_list
 
 @dataclass
 class Config:
+    csv: CsvConfig
     graphite: Optional[GraphiteConfig]
     grafana: Optional[GrafanaConfig]
     tests: Dict[str, TestConfig]
@@ -56,7 +58,9 @@ def load_templates(config: Dict) -> Dict[str, Dict]:
     return templates
 
 
-def load_tests(config: Dict, templates: Dict) -> Dict[str, TestConfig]:
+def load_tests(
+    config: Dict, templates: Dict, csv_config: Optional[CsvConfig] = None
+) -> Dict[str, TestConfig]:
     tests = config.get("tests", {})
     if not isinstance(tests, Dict):
         raise ConfigError("Property `tests` is not a dictionary")
@@ -71,7 +75,7 @@ def load_tests(config: Dict, templates: Dict) -> Dict[str, TestConfig]:
         except KeyError as e:
             raise ConfigError(f"Template {e.args[0]} referenced in test {test_name} not found")
         test_config = merge_dict_list(template_list + [test_config])
-        result[test_name] = create_test_config(test_name, test_config)
+        result[test_name] = create_test_config(test_name, test_config, csv_config)
 
     return result
 
@@ -98,13 +102,14 @@ def load_test_groups(config: Dict, tests: Dict[str, TestConfig]) -> Dict[str, Li
 
 
 def load_config_from_parser_args(args: configargparse.Namespace) -> Config:
+    csv_config = CsvConfig.from_parser_args(args)
     config_file = getattr(args, "config_file", None)
     if config_file is not None:
         yaml = YAML(typ="safe")
         config = yaml.load(Path(config_file).read_text())
 
         templates = load_templates(config)
-        tests = load_tests(config, templates)
+        tests = load_tests(config, templates, csv_config)
         groups = load_test_groups(config, tests)
     else:
         logging.warning("Otava configuration file not found or not specified")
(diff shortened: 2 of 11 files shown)


docs/CSV.md
```diff
@@ -34,9 +34,25 @@ tests:
     metrics: [metric1, metric2]
     csv_options:
       delimiter: ','
-      quotechar: "'"
+      quote_char: "'"
```
Gerrrr, line 37: Let's keep the existing name and not break existing configs unnecessarily?

Gerrrr (commented):
Thanks for the contribution! Please revert the breaking yaml config change, otherwise LGTM.

docs/CSV.md
MrlixiangWE: Good catch. I restored `quotechar` for the existing per-test YAML configuration and added regression coverage for it. The new top-level ConfigArgParse setting remains `csv.quote_char`, matching `--csv-quote-char` and `CSV_QUOTE_CHAR`.

docs/CSV.md
```diff
@@ -37,6 +37,22 @@ tests:
...
 ```
 
+## CSV options
+
+The delimiter and quote character can also be configured globally:
+
+```yaml
+csv:
+  delimiter: ';'
+  quote_char: "'"
```
Gerrrr, line 47: another one

MrlixiangWE: You're right, I missed the new top-level YAML example. I've changed it to `quotechar` too and added regression coverage for the top-level config path.

otava/config.py
```diff
@@ -128,13 +134,18 @@ class NestedYAMLConfigFileParser(configargparse.ConfigFileParser):
...
     CLI_CONFIG_SECTIONS = [
+        CsvConfig.NAME,
         GraphiteConfig.NAME,
         GrafanaConfig.NAME,
         SlackConfig.NAME,
         PostgresConfig.NAME,
         BigQueryConfig.NAME,
     ]
 
+    CONFIG_KEY_ALIASES = {
```
Gerrrr: I am really on the fence with `CONFIG_KEY_ALIASES`. @MrlixiangWE  can you please elaborate your thoughts on why do we want to introduce them?

My take is, if we really dislike `quotechar` - now is the time to break backwards compatibility and rename it to `quote-char` everywhere (although, I'd like to know the justification for it). OTOH if we can live with `quotechar`, why don't we just add `csv-quotechar` and call it a day?

MrlixiangWE: The alias table goes away either way — it was bridging a divergence this PR itself created. But let me give the justification you asked for, which I should have led with in the first round instead of just renaming things.

Across the config surface, multi-word names are separated: `time_column`, `csv_options`, `update_statement`, `test_groups`, `base_branch`, `bigquery.project_id` in YAML; `--bigquery-project-id`, `--since-commit`, `--config-file` on the CLI; `BIGQUERY_PROJECT_ID` and friends in the environment. `quotechar` is the one concatenated exception (not counting `hostname` and `username`, which are ordinary words). It comes from Python's csv API, but someone writing an otava config shouldn't need to know that — it's just one more special spelling to memorize. That's what the original rename was about — and it matched the names the code itself already used (`CsvOptions.quote_char`, and the per-test loader key). When you asked to keep the existing name, I took that to mean the per-test files specifically, restored `quotechar` there, and named the new flag `--csv-quote-char` to match the other options — then papered over the gap with the alias. In hindsight I should have raised the naming question openly instead of doing it in halves.

One thing worth knowing before we decide: on current master the per-test loader reads only `quote_char`:

https://github.com/apache/otava/blob/cd5bc1eedb2f4ecd851a76ef3211916774cb8afa/otava/test_config.py#L257-L259

so the `quotechar` in [docs/CSV.md](https://github.com/apache/otava/blob/cd5bc1eedb2f4ecd851a76ef3211916774cb8afa/docs/CSV.md?plain=1#L37) and [examples/csv/config/otava.yaml](https://github.com/apache/otava/blob/cd5bc1eedb2f4ecd851a76ef3211916774cb8afa/examples/csv/config/otava.yaml#L27) is silently ignored today — the example never notices because its CSV has no quoted fields. The docs and the code already disagree, and a full rename to `quote_char` wouldn't break any config that actually works.

So my vote is the deliberate rename: `quote_char` in YAML, `--csv-quote-char`, `CSV_QUOTE_CHAR`, docs and examples updated to match, no aliases and no fallback. If you'd rather keep `quotechar`, I'll do that instead — `--csv-quotechar` and done. Either way the PR ends with exactly one public spelling.

Gerrrr: I agree that since `quotechar` is ignored today, it makes sense to rename it. Let's do the following:
1. Remove `CONFIG_KEY_ALIASES`.
2. use `quote_char` as yaml config and `--csv-quote-char` for CLI, `CSV_QUOTE_CHAR` as env variable.

Please also rebase this PR against latest master - I had to update dependencies to fix CI.

MrlixiangWE: Done, all three:

1. `CONFIG_KEY_ALIASES` is removed.
2. `quote_char` is the YAML key; `--csv-quote-char` and `CSV_QUOTE_CHAR` are unchanged. The only `quotechar` left in the tree is the keyword argument of `csv.reader` itself, which we pass our `quote_char` value to.
3. Rebased onto latest master (47254d9).

198 non-container tests, ruff, isort, flake8 and the package build pass.

otava/csv_options.py
```diff
@@ -17,6 +17,39 @@
...
+class CsvConfig:
+    NAME = "csv"
+
+    delimiter: Optional[str] = None
+    quote_char: Optional[str] = None
+
+    @staticmethod
+    def add_parser_args(arg_group):
+        arg_group.add_argument(
+            "--csv-delimiter",
```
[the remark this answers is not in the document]
Gerrrr: Good feedback. I will merge this PR after this is fixed.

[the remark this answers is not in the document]
MrlixiangWE: Good catch — fixed. Both options now use a `single_character` type validator, so a multi-character value is rejected when the configuration is parsed:

    error: argument --csv-delimiter: must be a single character, got '::'

That covers the config file, the CLI flag and the environment variable, with tests for each. The per-test `csv_options` keys still accept multi-character values, but that behavior predates this PR — I can tighten it here too if you prefer.

Also rebased onto de46ddf.

Gerrrr (approved):
LGTM! Thank you for this contribution and for addressing feedback!

PR #169, opened 2026-08-13
Make AnalyzedSeries change points lazy properties

### LINKED ISSUE #79: AnalyzedSeries __compute_change_points() and __group_change_points_per_time() could be @properties instead
https://github.com/apache/otava/blob/8bb5e8eaaf8d0880650ad382c557963912e9454d/otava/series.py#L235C13-L235C15

I feel like it would be more natural python to have self.change_points, self.change_points_by_time be @properties. It would also save some computation in cases where the latter is never used, in which case it would never need to be computed.

MrlixiangWE:
Constructing an `AnalyzedSeries` runs change point detection for every metric up front, even when the caller never reads the results. #79 asks for `change_points` and `change_points_by_time` to become properties instead.

`change_points`, `weak_change_points`, `change_points_by_time` and `change_points_timestamp` are now read-only properties backed by private fields. The first access computes everything once and caches it; `append()` and `from_json()` write the backing fields directly. Two behavior notes:

- `change_points_timestamp` now records when the change points were actually computed rather than when the object was constructed. A timestamp stored in JSON still takes precedence on deserialization.
- Constructing with an explicit `change_points` argument used to leave `weak_change_points` unset, so reading it raised `AttributeError` unless `from_json` patched it afterwards. It now defaults to an empty collection.

Cost measured before (master, cd5bc1e) and after (this branch) with the same script on the same machine (Linux, Xeon @ 2.8GHz, 8 GB RAM, Python 3.11), seeded random data. Every number is the median of 11 runs and is copied verbatim from the millisecond output of the script below; six decimals in milliseconds is exact nanosecond resolution, so nothing is rounded away, and the script prints the raw nanosecond medians alongside:

| series | | construction | first access | repeated access |
|--|--|--|--|--|
| 1 metric x 2000 points | before | 68.541473 ms | 0.000726 ms | 0.000298 ms |
| | after | 0.053550 ms | 65.670122 ms | 0.001036 ms |
| 4 metrics x 5000 points | before | 640.987496 ms | 0.000765 ms | 0.000234 ms |
| | after | 0.439367 ms | 623.203122 ms | 0.000977 ms |

The cost moves from the constructor to the first read; the total for callers that do read change points stays in the same range, and repeated reads stay cached. Detection results are unchanged: no expected value in the test suite needed to change. The only adjusted test is `test_validate`, which used to fake an uncomputed state by assigning `change_points = None` and now constructs with an explicitly empty collection.

<details>
<summary>Benchmark script</summary>

```python
import random
import statistics
import time

from otava.series import AnalysisOptions, AnalyzedSeries, Metric, Series

REPS = 11


def build_series(n, m):
    random.seed(42)
    metrics = {f"m{i}": Metric(1, 1.0) for i in range(m)}
    data = {f"m{i}": [random.gauss(100, 5) for _ in range(n)] for i in range(m)}
    return Series("bench", None, list(range(n)), metrics, data, {})


def run(n, m):
    construct, first, repeat = [], [], []
    for _ in range(REPS):
        s = build_series(n, m)
        t0 = time.perf_counter_ns()
        a = AnalyzedSeries(s, AnalysisOptions())
        t1 = time.perf_counter_ns()
        _ = a.change_points
        _ = a.change_points_by_time
        t2 = time.perf_counter_ns()
        _ = a.change_points
        _ = a.change_points_by_time
        t3 = time.perf_counter_ns()
        construct.append(t1 - t0)
        first.append(t2 - t1)
        repeat.append(t3 - t2)

    def fmt(v):
        med = statistics.median(v)
        return f"{med / 1e6:.6f} ms ({med:.0f} ns)"

    print(f"{m} metrics x {n} points")
    print(f"  construction:    {fmt(construct)}")
    print(f"  first access:    {fmt(first)}")
    print(f"  repeated access: {fmt(repeat)}")


run(2000, 1)
run(5000, 4)

import otava.series
print(f"module: {otava.series.__file__}")
```

</details>

Verification:

- `pytest tests perf` (non-container): 194 passed, including three new regression tests for laziness/caching, `append()` on a freshly constructed instance, and deserialization without recomputation
- `flake8`, `ruff check`, `ruff format --check`, `isort --check-only`: clean
- `git diff --check`: clean

Closes #79

### DIFF
--- a/otava/series.py
+++ b/otava/series.py
@@ -114,16 +114,12 @@ class Series:
     def analyze(self, options: Optional[AnalysisOptions] = None) -> "AnalyzedSeries":
         if options is None:
             options = AnalysisOptions()
-        logging.info(f"Computing change points for test {self.test_name}...")
         return AnalyzedSeries(self, options)
 
 
 class AnalyzedSeries:
     """
     Time series data with computed change points.
-
-    Change points are computed lazily, on first access. Constructing an
-    instance is cheap for callers that never read them.
     """
 
     __series: Series
@@ -144,6 +140,7 @@ class AnalyzedSeries:
 
     def __ensure_change_points_computed(self):
         if self.__change_points is None:
+            logging.info(f"Computing change points for test {self.__series.test_name}...")
             cp, weak_cps = self.__compute_change_points(self.__series, self.options)
             self.__change_points = cp
             self.__weak_change_points = weak_cps
@@ -254,8 +251,8 @@ class AnalyzedSeries:
         return self._validate_append(time, new_data, attributes) is None
 
     def _validate_append(self, time, new_data, attributes):
-        if not self.change_points:
-            return RuntimeError("You must use __compute_change_points() once first.")
+        # appending updates the cached results, so they must exist first
+        self.__ensure_change_points_computed()
         if not isinstance(time, list):
             return ValueError("time argument must be an array.")
         if not isinstance(new_data, dict):
@@ -300,13 +297,19 @@ class AnalyzedSeries:
 
         for metric in self.__series.data.keys():
             if metric not in new_data:
-                weak_change_points[metric] = self.weak_change_points.select_metrics(metric)
+                if metric in self.weak_change_points:
+                    weak_change_points[metric] = self.weak_change_points.select_metrics(metric)
                 continue
 
             new_data_len = len(new_data[metric])
+            previous_weak_cp = (
+                self.weak_change_points.get_change_points_for_metric(metric)
+                if metric in self.weak_change_points
+                else []
+            )
             old_weak_cp = [
                 cp
-                for cp in self.weak_change_points.get_change_points_for_metric(metric)
+                for cp in previous_weak_cp
                 if cp.index < len(self.__series.data[metric]) - new_data_len - 1
             ]
             change_points, weak_cps = compute_change_points(
@@ -348,7 +351,9 @@ class AnalyzedSeries:
         for metric, cpglist in r.items():
             self.change_points[metric] = cpglist
         self.__weak_change_points = w
-        self.__change_points_by_time = self.change_points.by_time()
+        # invalidate rather than rebuild: the property recomputes it on first read
+        self.__change_points_by_time = None
+        self.__change_points_timestamp = datetime.now(timezone.utc)
         return r, w
 
     def test_name(self) -> str:
--- a/otava/series.py
+++ b/otava/series.py
@@ -121,29 +121,56 @@ class Series:
 class AnalyzedSeries:
     """
     Time series data with computed change points.
+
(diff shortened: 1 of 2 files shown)


otava/series.py
```diff
@@ -132,29 +132,56 @@ def analyze(self, options: AnalysisOptions = AnalysisOptions()) -> "AnalyzedSeri
 class AnalyzedSeries:
     """
     Time series data with computed change points.
+
+    Change points are computed lazily, on first access. Constructing an
```
Gerrrr: nit: I don't think we need this addition to the doc.

tests/series_test.py
```diff
@@ -407,3 +408,83 @@ def test_series_raw_initialization():
...
+        "append_lazy_test",
+        branch=None,
+        time=list(range(len(data))),
+        metrics={"m1": Metric(1, 1.0)},
+        data={"m1": data},
+        attributes={},
+    )
+
+    analyzed = test.analyze()
+    analyzed.append(time=[len(data)], new_data={"m1": [5.0]}, attributes={})
```
Gerrrr, line 742: This only covers a series with an existing change point. A stable series computes an empty result, which `_validate_append()` mistakes for “not computed.” Could you please add a case that does not produce a change point and explicitly ensure computation rather than checking collection truthiness?

otava/series.py
```diff
@@ -132,29 +132,56 @@ def analyze(self, options: AnalysisOptions = AnalysisOptions()) -> "AnalyzedSeri
...
+        self.__weak_change_points = ChangePointsByMetric() if change_points is not None else None
+        self.__change_points_by_time = None
+        # records when the change points were calculated
+        self.__change_points_timestamp = (
+            datetime.now(timezone.utc) if change_points is not None else None
+        )
+
+    def __ensure_change_points_computed(self):
+        if self.__change_points is None:
+            cp, weak_cps = self.__compute_change_points(self.__series, self.options)
```
Gerrrr, line 144: `logging.info("Computing change points...` from line 128 should move here.

Gerrrr (commented):
This is a very nice optimization, thank you for opening this PR! I left a few suggestions + please rebase against latest master.

otava/series.py
```diff
@@ -320,8 +350,9 @@ def append(self, time, new_data, attributes):
         # r has a subset of all metrics, so can't just set change_points to r
         for metric, cpglist in r.items():
             self.change_points[metric] = cpglist
-        self.weak_change_points = w
-        self.change_points_by_time = self.change_points.by_time()
+        self.__weak_change_points = w
+        self.__change_points_by_time = self.change_points.by_time()
```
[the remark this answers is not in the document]
MrlixiangWE: Fixed

Gerrrr (approved):
LGTM, thanks!

PR #170, opened 2026-08-16
Add InfluxDB importer

adambernier:
Add InfluxDB 3 configuration, query importing, Arrow result normalization, tests,
    documentation, and CLI help snapshots.

### DIFF
--- a/docs/INFLUXDB.md
+++ b/docs/INFLUXDB.md
@@ -61,7 +61,7 @@ object storage for production deployments.
 
 ```yaml
 tests:
-  api_latency:
+  api_latency_sql:
     type: influxdb
     query_language: sql
     query: |
--- a/docs/INFLUXDB.md
+++ b/docs/INFLUXDB.md
@@ -1,0 +1,80 @@
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License. You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ -->
+
+# Importing results from InfluxDB 3
+
+Otava imports query results from InfluxDB 3 Core or Enterprise through the
+[`influxdb3-python`](https://docs.influxdata.com/influxdb3/core/reference/client-libraries/v3/python/)
+client. SQL is the default query language; set `query_language: influxql` for
+InfluxQL queries.
+
+## Connection
+
+```yaml
+influxdb:
+  host: http://localhost:8181
+  database: performance
+  token: ${INFLUXDB_TOKEN}
+```
+
+The same settings are available through `INFLUXDB_HOST`, `INFLUXDB_DATABASE`,
+and `INFLUXDB_TOKEN`, or the `--influxdb-host`, `--influxdb-database`, and
+`--influxdb-token` command-line options. Command-line values take precedence
+over environment variables, which take precedence over YAML.
+
+## Test configuration
+
+```yaml
+tests:
+  api_latency:
+    type: influxdb
+    query_language: sql
+    query: |
+      SELECT time, branch, p95_ms, commit
+      FROM api_latency
+      WHERE branch = %{BRANCH}
+      ORDER BY time
+    time_column: time
+    attributes: [branch, commit]
+    metrics:
+      p95:
+        column: p95_ms
+        direction: -1
+        scale: 1
+
+  legacy_api_latency:
+    type: influxdb
+    query_language: influxql
+    query: SELECT time, branch, p95_ms FROM api_latency WHERE branch = %{BRANCH}
+    attributes: [branch]
+    metrics: [p95_ms]
+```
+
+Metric definitions use `column`, `direction`, and `scale` as with the other
+SQL-backed importers. `%{BRANCH}` is replaced with an escaped string literal
+when `--branch` is supplied.
+
+Run the analysis with:
+
+```bash
+otava analyze api_latency --branch main --last 100
+```
+
+InfluxDB is import-only in this release; Otava does not write change points
+back to InfluxDB.
--- a/docs/INFLUXDB.md
+++ b/docs/INFLUXDB.md
@@ -38,6 +38,25 @@ and `INFLUXDB_TOKEN`, or the `--influxdb-host`, `--influxdb-database`, and
 `--influxdb-token` command-line options. Command-line values take precedence
 over environment variables, which take precedence over YAML.
 
+## Reproducible example
+
+The bundled example starts InfluxDB 3 Core with authenticated, in-memory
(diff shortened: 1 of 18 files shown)


otava/importer.py
```diff
@@ -827,6 +830,90 @@ def fetch_all_metric_names(self, test_conf: BigQueryTestConfig) -> List[str]:
...
+            metric_indexes = [columns.index(metric.column) for metric in metrics.values()]
+        except ValueError as err:
+            raise DataImportError(f"Column not found {err.args[0]}")
+
+        time = []
+        data = {name: [] for name in metric_names}
+        attributes = {columns[index]: [] for index in attr_indexes}
+        for row in rows:
+            timestamp = row[time_index]
+            if timestamp < since_time or timestamp >= until_time:
```
Gerrrr, line 902: InfluxDB returns time as Arrow `Timestamp(Nanosecond, None)`. Converting a real InfluxDB result to Python produces a timezone-naive datetime, while Otava's time bounds are timezone-aware:

```python
from datetime import datetime
import pyarrow
from otava.data_selector import DataSelector

table = pyarrow.table(
    {
        "time": pyarrow.array(
            [datetime(2024, 1, 2)],
            type=pyarrow.timestamp("ns"),
        )
    }
)

influxdb_time = table.to_pylist()[0]["time"]
selector = DataSelector()

influxdb_time < selector.until_time
```

This throws:
```python
TypeError: can't compare offset-naive and offset-aware datetimes
```

This means the importer fails while filtering normal InfluxDB results.

We should normalize InfluxDB timestamps to UTC before comparing them and calling .timestamp(). Please add a test with an Arrow timestamp[ns] column without a timezone. The current test data explicitly uses `timezone.utc`, which hides this problem.

tests/influxdb_test.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+    assert test.query_language == "influxql"
+    with pytest.raises(TestConfigError):
+        create_test_config(
+            "latency",
+            {"type": "influxdb", "query": "SELECT * FROM latency", "metrics": ["p95_ms"], "query_language": "flux"},
+        )
+
+
+def test_influxdb_importer_reads_arrow_table_and_applies_selection():
```
Gerrrr, line 100: We should have an e2e test with a Dockerized InfluxDB instance, similar to `postgres_e2e_test.py` and `graphite_e2e_test.py`.

Mocking the returned Arrow table does not test the actual client/server contract, authentication, query language, or returned schema. In particular, an e2e test would have caught the timezone issue above.

examples/influxdb/otava.yaml
```diff
@@ -0,0 +1,23 @@
+# InfluxDB 3 connection settings can also be supplied with INFLUXDB_HOST,
```
Gerrrr, line 18: Please make this a reproducible example. PostgreSQL and Graphite examples include Docker Compose configuration, seeded data, and an Otava configuration. This directory only contains otava.yaml, so there is no way to run the example as documented.

Gerrrr (commented):
Thanks for the contribution! This PR looks good overall, but needs a few changes to get ready for merge. Please also rebase against latest master - I had to update dependencies to fix CI.

otava/importer.py
```diff
@@ -827,6 +830,90 @@ def fetch_all_metric_names(self, test_conf: BigQueryTestConfig) -> List[str]:
...
+            )
+
+        metrics = self.__selected_metrics(test_conf.metrics, selector.metrics)
+        query = test_conf.query
+        if "%{BRANCH}" in query:
+            if not selector.branch:
+                raise DataImportError(
+                    f"Test {test_conf.name} uses %{{BRANCH}} in query but --branch was not specified"
+                )
+            branch_literal = "'" + selector.branch.replace("'", "''") + "'"
```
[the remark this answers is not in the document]
adambernier: Fixed in 0e5b49f. SQL branch literals keep doubled-quote escaping, while InfluxQL now escapes backslashes, quotes, carriage returns, and newlines in the required order. I split the regression coverage by query language and assert the exact generated queries.

otava/importer.py
adambernier: Fixed in 0e5b49f. Timezone-less Arrow timestamps are now interpreted as UTC before range filtering and epoch conversion. The Arrow-backed test uses an explicit timestamp(ns) column without a timezone and asserts both filtering and UTC epoch values; timezone-aware rows remain covered by the query-language tests.

tests/influxdb_test.py
adambernier: Fixed in 8757974. I added a Docker-backed InfluxDB 3 Core E2E test using the pinned 3.11.2-core image, a preconfigured offline admin token, memory storage, and a random host port. It runs the shared seed script, loads the bundled SQL and InfluxQL test definitions through Otava, and asserts identical ordered UTC timestamps, branch/commit attributes, metrics, and values from the real authenticated server. The new test and the full 204-test suite pass locally.

examples/influxdb/otava.yaml
adambernier: Fixed in 8757974. The InfluxDB example now includes Compose services for the pinned authenticated server, one-shot seeding, and Otava; a fixed demonstration-only offline admin token; deterministic line protocol with timestamps, branches, commits, and latency values; and an idempotent readiness-aware seed script. The Otava config has equivalent SQL and InfluxQL tests, and the guide documents the build/Compose/analyze workflow plus the test-only credential and ephemeral-memory caveats. I manually exercised both query-language commands against the Compose stack.

tests/influxdb_e2e_test.py
```diff
@@ -0,0 +1,109 @@
...
+        config = load_config_from_file(
+            str(EXAMPLE_DIR / "otava.yaml"),
+            arg_overrides=[
+                "--influxdb-host",
+                host,
+                "--influxdb-token",
+                INFLUXDB_TOKEN,
+            ],
+        )
+        importer = InfluxDBImporter(InfluxDB(config.influxdb))
```
Gerrrr: This test starts a real InfluxDB server, but it bypasses Otava’s public execution path by constructing InfluxDBImporter directly. As a result, it would still pass if the CLI options, configuration wiring,
`Importers.get()` registration, analysis, or reporting were broken. Could this invoke `otava analyze`, as the PostgreSQL and Graphite E2E tests do, and assert the resulting output?

docs/INFLUXDB.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+Metric definitions use `column`, `direction`, and `scale` as with the other
+SQL-backed importers. `%{BRANCH}` is replaced with an escaped string literal
+when `--branch` is supplied.
+
+Run the analysis with:
+
+```bash
+otava analyze api_latency_sql --branch main --last 100
```
Gerrrr, line 95: The configuration above defines the test as `api_latency`, but this command runs `api_latency_sql`. Let's rename `api_latency` above into `api_latency_sql` to make it consistent with the rest of the doc and the example.

Gerrrr (commented):
2 more minor changes + fix rebase conflict and it will be ready to merge.

tests/influxdb_e2e_test.py
adambernier: Fixed in 7c47ab5. The Docker-backed E2E test now runs uv run otava analyze for both api_latency_sql and api_latency_influxql through the public CLI/configuration path. It supplies the random host URL, database, token, branch, config path, and explicit UTC bounds, then normalizes and compares both stdout values to the same exact six-row table including timestamps, branch/commit attributes, p95 values, and the +37.2% detected change. Rich subprocess diagnostics are retained.

docs/INFLUXDB.md
adambernier: Fixed in 7c47ab5. The documentation snippet now names the SQL test api_latency_sql, matching both the bundled example and the analysis command.

Gerrrr (approved):

PR #178, opened 2026-09-01
Resolve #78: migrate domain models to Pydantic v2

### LINKED ISSUE #78: Use pydantic classes for json serialization
The Series and AnalyzedSeries classes support serializing themselves into json, via series.as_json() calls. This is not pythonic at all. (https://github.com/apache/otava/blob/8bb5e8eaaf8d0880650ad382c557963912e9454d/otava/series.py#L109)

It seems the modern best practice is to use Pydantic classes, which are like @dataclasses but better. In particular, a pydantic object can just be cast to dict(obj) or alternatively to json. https://docs.pydantic.dev/latest/concepts/serialization/

adambernier:
Resolves #78.

  Migrates domain models to Pydantic v2 while preserving
  persistence and report JSON compatibility.


docs/BASICS.md
```diff
@@ -19,6 +19,21 @@
 
 # Basics
 
+## Python persistence API
```
Gerrrr: Otava does not officially support/document Python API yet, so I would probably skip this section. 

In general, the plan is to keep 0.8.* as the stable release branch and iterate on API, distribution, features, etc 0.9.*+. Once we stabilize the APIs, we'll release 1.0.0.

PR #179, opened 2026-09-04
Define the public API in otava/__init__.py

### LINKED ISSUE #101: Documentation on using Otava Programmically.
I recently utilized Otava by calling the functions directly instead of using an importer.  This is useful if you have data received by Python in some custom way and want to programmically utilize Otava on the data.  There isn't a lot of documentation on this so it was sort of like looking at unit tests and guessing at it.  I figured it out, but I am not sure if I am doing it entirely correctly.  It would be useful if this was documented officially as an option, unless for some reason you rather people be limited to using an existing importer?  

Here is my potentially bad attempt at an example of this

```
from otava.analysis import compute_change_points, ChangePoint`

def _compute_change_points(p, m=0.0, w=30, series=[], new_data=None):
    return compute_change_points(series, window_len=w, max_pvalue=p, min_magnitude=m, new_data=new_data)

#Bring in your data from another source, and set y_list to the data...
y_list = []

#Tweak the parameters to appropriate max_pvalue and min_magnitude options, as these may differ depending on dataset and sensitivity desired


math_results,old_math_results = _compute_change_points(0.005, m=0.003, w=100, series=y_list, new_data=None)
index_list = []

c: ChangePoint
for c in math_results:
    index_list.append(c.index)

for index, y in enumerate(y_list):
    if index in index_list:
        print("This Y should be marked!")

### PR DESCRIPTION
**Pull request by MrlixiangWE**
Expose nine names from `otava`, including `Series`, `AnalysisOptions` and
`compute_change_points`, and document their use in `docs/API.md`. Existing
submodule imports remain available.

The package root now loads SciPy, so importing a lightweight module such as
`otava.csv_options` also loads those dependencies.
Median import overhead in milliseconds, measured over 21 runs per checkout,
alternating base and branch, on a 4-core QEMU/KVM VM with Python 3.12.14.
Each sample subtracts a separate Python startup measurement, which can yield
negative estimates.

| Import | `d2d51c9` | `610066f` |
| --- | ---: | ---: |
| `otava` | -1 | 593 |
| `otava.csv_options` | 17 | 581 |
| `otava.util` | 168 | 818 |
| `otava.series` | 598 | 601 |
| `otava.main` | 842 | 866 |

The tests check the documented export names against `__all__`, check option
names and defaults against `AnalysisOptions()`, and run the documentation
examples. The wheel smoke test checks the installed package's exports and
calls `compute_change_points` without extras.

Verified on Ubuntu 22.04 / x86_64 with `uv sync --locked --all-extras --dev`:
248 tests passed on each of Python 3.10.21, 3.11.16, 3.12.14, 3.13.15 and
3.14.7; the same suite passed 241 tests on `d2d51c9`. This covers `tests` and
`perf`, excluding the Graphite, PostgreSQL and InfluxDB container e2e modules.
On Python 3.12.14, `sphinx-build -W --keep-going` passed on both checkouts.
A wheel built from this branch passed `tests/core_install_smoke.py` in an
installation without extras.

Closes #101

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1,0 +1,102 @@
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ -->
+
+# Using Otava as a library
+
+Otava can be used without its command line interface and without any of its
+importers: build a `Series` from data you already have, analyze it, and read the
+change points back.
+
+```python
+from otava import AnalysisOptions, Metric, Series
+
+# One measurement per commit. Metric.direction says which way is an improvement:
+# 1 when higher is better, -1 when lower is better.
+series = Series(
+    test_name="throughput",
+    branch=None,
+    time=list(range(60)),
+    metrics={"ops": Metric(direction=1, scale=1.0)},
+    data={"ops": [100.0] * 30 + [80.0] * 30},
+    attributes={"commit": [f"sha{i}" for i in range(60)]},
+)
+
+analyzed = series.analyze(AnalysisOptions(window_len=20, max_pvalue=0.001))
+
+for group in analyzed.change_points_by_time:
+    for metric, change in group.changes.items():
+        print(
+            group.attributes["commit"],
+            metric,
+            change.stats.mean_before(),
+            change.stats.mean_after(),
+            f"{change.stats.forward_change_percent():+.1f}%",
+        )
+```
+
+`analyze()` does not interpret `time`, so a plain index works for data that is
+not a time series, but keep it increasing: change points are collected in time
+order, and a second group that is not later than the first raises `ValueError`.
+Otava's own reports do interpret it, formatting it with
+`datetime.fromtimestamp`, so anything printed by Otava reads it as seconds since
+the epoch.
+
+`attributes` carries one value per data point, and the values at a change point
+are returned on its group. That is how you get from a detected change back to
+the commit that caused it.
+
+`Series.analyze()` is lazy: the change points are computed the first time any of
+the change point properties on `AnalyzedSeries` is read, and cached from then on.
+Errors the analysis raises therefore surface at that first read.
+
+## What is public
+
+The names in `otava.__all__`, re-exported from the package root:
+
+| Name | Purpose |
+| --- | --- |
+| `Series` | Input data: timestamps, metrics, values, per-point attributes |
+| `Metric` | Direction, scale and unit of one metric |
+| `AnalysisOptions` | `window_len`, `max_pvalue`, `min_magnitude`, `orig_edivisive` |
+| `AnalyzedSeries` | Result of `Series.analyze()` |
+| `ChangePoint` | One change in one metric; the statistics are on its `stats` |
+| `ChangePointGroup` | Changes that share a point in time |
+| `ChangePointsByMetric` | Change points keyed by metric |
+| `ChangePointsByTime` | Change point groups in time order |
+| `compute_change_points` | The detection algorithm on a bare sequence of floats; returns `(strong, weak)` |
+
+Everything else is internal, including every submodule path such as
+`otava.importer`, `otava.config` and `otava.main`. Otava does not offer a stable
+public API yet, so even the names above can still change in a minor or patch
+release.
+
+## Options
+
+`AnalysisOptions` holds the settings the `analyze` command exposes as flags,
+with the same defaults. The attribute names and the flag names differ; run
+`otava analyze --help` for the flags.
+
+| Option | Default | Effect |
+| --- | --- | --- |
+| `window_len` | 50 | Number of points the algorithm looks at around a candidate |
+| `max_pvalue` | 0.001 | Significance threshold a change point must meet |
+| `min_magnitude` | 0.0 | Smallest relative change worth reporting |
+| `orig_edivisive` | False | Use unmodified e-divisive instead of the windowed variant |
+
+See [Change Point Detection](MATH.md) for what these do to the results.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: docs/API.md [hunk 1]
```diff
@@ -0,0 +1,102 @@
...
+
+Otava can be used without its command line interface and without any of its
+importers: build a `Series` from data you already have, analyze it, and read the
+change points back.
+
+```python
+from otava import AnalysisOptions, Metric, Series
+
+# One measurement per commit. Metric.direction says which way is an improvement:
+# 1 when higher is better, -1 when lower is better.
```
henrikingo: It would be good to also define constants like Metric.HIGHER_IS_BETTER and Metric.LOWER_IS_BETTER

docs/API.md
```diff
@@ -0,0 +1,102 @@
...
+series = Series(
+    test_name="throughput",
```
henrikingo: throughput is also a metric, of which operations per second is the unit how it's measured.

test_name could be just send_hello_message for example. start_engines. Etc...

docs/API.md
```diff
@@ -0,0 +1,102 @@
...
+    branch=None,
+    time=list(range(60)),
+    metrics={"ops": Metric(direction=1, scale=1.0)},
+    data={"ops": [100.0] * 30 + [80.0] * 30},
+    attributes={"commit": [f"sha{i}" for i in range(60)]},
+)
+
+analyzed = series.analyze(AnalysisOptions(window_len=20, max_pvalue=0.001))
+
+for group in analyzed.change_points_by_time:
```
henrikingo: Personally I'm not very fond of the name GhangePointGroup to describe a set of change points at the same timestamp. Essentially it's a row. It is what it is but let's not propagate the use of "group" further. 

My suggestion, call the variable `changes_at` , `change_points_at` or `change_points_at_time`. (cp_at would also be a good name but experience has taught me that emphasizing plural vs singular can help sometimes)

docs/API.md
```diff
@@ -0,0 +1,102 @@
...
+            f"{change.stats.forward_change_percent():+.1f}%",
+        )
+```
+
+`analyze()` does not interpret `time`, so a plain index works for data that is
+not a time series, but keep it increasing: change points are collected in time
+order, and a second group that is not later than the first raises `ValueError`.
+Otava's own reports do interpret it, formatting it with
+`datetime.fromtimestamp`, so anything printed by Otava reads it as seconds since
+the epoch.
```
henrikingo: This is btw a perfect time to discuss whether we still want to use unix timestamps, datetime() or possibly allow both?
Personal opinion: At this level the unix ts is good I guess, but when serialized as JSON or CSV I've almost started thingking a human readable YYYY-MM-DDThh:mm:ss.dddZ` string is more convenient when browsing data.

docs/API.md
```diff
@@ -0,0 +1,102 @@
...
+`AnalysisOptions` holds the settings the `analyze` command exposes as flags,
+with the same defaults. The attribute names and the flag names differ; run
+`otava analyze --help` for the flags.
+
+| Option | Default | Effect |
+| --- | --- | --- |
+| `window_len` | 50 | Number of points the algorithm looks at around a candidate |
+| `max_pvalue` | 0.001 | Significance threshold a change point must meet |
+| `min_magnitude` | 0.0 | Smallest relative change worth reporting |
+| `orig_edivisive` | False | Use unmodified e-divisive instead of the windowed variant |
```
henrikingo: See #154 which changes this last one. It would be nice to merge it before this API, but of course you should  work from what is in trunk now. So this is just fyi.

When using orig_edivisive or the new deterministic edivisive, window_len is not used at all and it is my opinion that usually min_magnitude should also not be used. In other words max_pvalue would be the only parameter you need to fiddle with.

docs/API.md
```diff
@@ -0,0 +1,102 @@
...
+| `Series` | Input data: timestamps, metrics, values, per-point attributes |
+| `Metric` | Direction, scale and unit of one metric |
+| `AnalysisOptions` | `window_len`, `max_pvalue`, `min_magnitude`, `orig_edivisive` |
+| `AnalyzedSeries` | Result of `Series.analyze()` |
+| `ChangePoint` | One change in one metric; the statistics are on its `stats` |
+| `ChangePointGroup` | Changes that share a point in time |
+| `ChangePointsByMetric` | Change points keyed by metric |
+| `ChangePointsByTime` | Change point groups in time order |
+| `compute_change_points` | The detection algorithm on a bare sequence of floats; returns `(strong, weak)` |
+
```
henrikingo: This would really benefit from a graph that shows the relationships of the classes: Metric are a property of Series, AnalysisOptions a parameter to analyze(), which returns AnalyzedSeries. And all the ChangePoint* are containers of each other, and ultimately property of AnalyzedSeries.

`compute_change_points` feels off here. Why is it needed?

otava/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+    >>> series = Series(
+    ...     "throughput",
+    ...     branch=None,
+    ...     time=list(range(40)),
+    ...     metrics={"ops": Metric(direction=1, scale=1.0)},
+    ...     data={"ops": [100.0] * 20 + [80.0] * 20},
+    ...     attributes={},
+    ... )
+    >>> analyzed = series.analyze(AnalysisOptions(window_len=20))
+    >>> [group.time for group in analyzed.change_points_by_time]
```
henrikingo: Same comment for the `group` variable as above

tests/public_api_test.py
```diff
@@ -0,0 +1,184 @@
...
+def optional_client_modules():
+    """Top-level modules the optional service clients import.
+
+    Read off the import_optional_dependency() call sites rather than repeated
+    here, so that a new optional client is covered the day it is added.
+    """
+    call_site = re.compile(r"""import_optional_dependency\(\s*["']([^"']+)["']""")
+    names = {
+        match.group(1).partition(".")[0]
+        for path in (REPO_ROOT / "otava").glob("*.py")
```
henrikingo: Ok I'm sure this works in practice but wouldn't  .pyc and .ipy and so on be valid suffixes? In general it feels iffy to depend on even he idea that a file system is involved...

tests/public_api_test.py
```diff
@@ -0,0 +1,184 @@
...
+    would silently make `title` the only section allowed to come last.
+    """
+    body = re.search(
+        rf"^## {re.escape(title)}\n(.*?)(?=^## |\Z)",
+        DOCS_API.read_text(encoding="utf-8"),
+        re.M | re.S,
+    )
+
+    assert body, f"no '{title}' section in {DOCS_API}"
+    return body.group(1)
```
henrikingo: I wonder if its necessary to test for the existence of docs?

henrikingo:
Thanks MirliangWE! I hope to review this soon.Defining  an API for Otava use in a library is a valuable addition.

MrlixiangWE:
Thanks, Henrik! I’ve also pushed a small follow-up to clarify the documentation around timestamp handling and the supported import surface.

